PR #10, opened 2023-08-28
PREOPS-3639: support setting custom prenight tabs from a file specified on the command line

ehneilsen:
These changes should also close PREOPS-3600.

### DIFF
--- a/notebooks/custom_plot.json
+++ b/notebooks/custom_plot.json
@@ -1,0 +1,23 @@
+[
+    {
+        "name": "Seeing",
+        "settings": {
+            "kind": "scatter",
+            "x": "start_date",
+            "y": [
+                "seeingFwhmEff",
+                "seeingFwhm500",
+                "seeingFwhmGeom"
+            ]
+        }
+    },
+    {
+        "name": "Sky brightness",
+        "settings": {
+            "kind": "hist",
+            "y": [
+                "skyBrightness"
+            ]
+        }
+    }
+]
--- a/notebooks/prenight.ipynb
+++ b/notebooks/prenight.ipynb
@@ -71,6 +71,9 @@
     "from pytz import timezone\n",
     "import lzma\n",
     "import pickle\n",
+    "import yaml\n",
+    "import json\n",
+    "from collections import OrderedDict\n",
     "from tempfile import TemporaryDirectory, NamedTemporaryFile"
    ]
   },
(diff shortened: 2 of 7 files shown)


rhiannonlynne:
It looks good in a notebook and I can add a new custom plot.
However, I must be doing something wrong from the command line. 
I took the same opsim output and pickle from the notebook, and tried to start up the app. 
In the notebooks directory, I just entered 
`python ../schedview/app/prenight/prenight.py`
and then pointed it at my test_opsim.db and test_scheduler.pickle.xz files, and changed the date to the first date of the simulation. 
After changing the date a few times, the airmass histogram plot showed up in the app .. but I couldn't change to the other tabs. I'm not entirely sure if this is "the app got confused because I changed the date and the names of the files" or something else. 
Happy to demo if it helps, but it may be that I just got the invocation of the app incorrect?

rhiannonlynne:
Done (and demo'd with Eric my problem). 
Eric will check on USDF. 
Maybe a package version issue?

rhiannonlynne:
I don't know if that was the difference or if it was rebooting my machine (which removed that stale Jupyter notebook server which wasn't actually running) but this works now even with custom tabs using 
`python ../../schedview/app/prenight/prenight.py --night 2025-01-01 --opsim_db test_opsim.db --scheduler test_scheduler.pickle.xz -c ../test_custom_tabs.json` 

That said, I think there's something still a bit off about loading up the visits from the right night if you change it .. 
when I started the app with the wrong night on purpose (putting in --night 2025-01-03), I got the blank plots and "No visits on this night message" in the skymap + visit table tabs as expected --but, if I changed the night on the widget to 2025-01-01, then I still had blank plots and "No visits on this night" messages.

ehneilsen:
I haven't been able to reproduce this with your example files. I did find and fix a different bug, though: after changing the date, depending on the size and aspect ratio of your browser window, the horizon coordinates plot would sometimes only show half of the sky after the change of date. I suspect this is actually a panel bug, but I did find a way to work around it.
Also, I haven't quite been able to exactly test with the files you've provided, because the version of rubin_sim I'm using (which is the latest on main) can't load the scheduler pickle you're using. But, the only thing this should affect is the uranography plot, so I doubt if that's the issue.

rhiannonlynne:
So .. I tried it again today. I *think* what might have been happening is that it's just really slow. 
When I tried it today, I was distracted by another task and so left the dashboard alone for several minutes, and when I did come back, it was updated. I think I did find that figure sizing issue you mentioned though.

rhiannonlynne (approved):

PR #11, opened 2023-09-01
Tickets/preops 3725: Keep size and aspect ration of horizon plot reasonable when date changed

ehneilsen:


rhiannonlynne:
Something on this ticket seems to stop the automatic pop-up of the app in safari (i.e. I run the python prenight.py command and then have to open the link in safari myself - at which point the rest of the schedview stuff loads). 

Otherwise, yes, the alt/az plot now maintains its appropriate sizing even when no visits are present. 

It takes over a minute to update the display when I change nights. Is this expected?

rhiannonlynne (approved):

PR #16, opened 2023-09-08
PREOPS-3732:  only load 1 night of pre-computed sky brightness

ehneilsen:
The first of a few optimizations


rhiannonlynne (approved):
Definitely a bit snappier now.

PR #17, opened 2023-09-13
Tickets/PREOPS-3643: Fix ruff & doc format issues in schedview and implement standard git pre-commits.

ehneilsen:
`ruff`, `black`, and `isort` are all in the pre-commit, and applied to the code.
The docs layout now matches that of `uranography`, but don't contain much beyond what sphinx produces.
Expanding the docs content and push to a website should probably be in a different ticket; this is a big enough commit as it is.


pyproject.toml
```diff
@@ -31,3 +31,51 @@ __version__ = "{version}"
...
+    "E",  # pycodestyle
+    "F",  # pyflakes
+    "N",  # pep8-naming
+    "W",  # pycodestyle
+    #"D",  # pydocstyle
+]
+target-version = "py310"
+extend-select = [
+    "RUF100", # Warn about unused noqa
+]
```
rhiannonlynne, line 75: This doesn't seem to match the ruff configuration at https://developer.lsst.io/python/style.html

schedview/compute/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    compute_basis_function_reward_at_time,
+    compute_basis_function_rewards,
+    create_example,
+    make_scheduler_summary_df,
+    make_unique_survey_name,
+    replay_visits,
+)
+from .survey import compute_maps, make_survey_reward_df
+
+__all__ = [
```
rhiannonlynne, line 13: When John reformatted some of rubin_sim, he moved the _all_ ahead of the imports and used a tuple. 
I think this matches the DM guidelines - https://developer.lsst.io/python/style.html#standard-code-order-should-be-followed - so since this a PR with a lot of reformatting to match standards goals, could you please also move the __all__ statements accordingly?

schedview/compute/maf_metrics.py
```diff
@@ -1,12 +1,13 @@
...
-import schedview
-import schedview.compute.astro
+from rubin_sim.maf import MetricBundleGroup  # noqa F401
 
 # Imported to help sphinx make the link
 from rubin_sim.scheduler.model_observatory import ModelObservatory  # noqa F401
-from rubin_sim.maf import MetricBundleGroup  # noqa F401
+
+import schedview
+import schedview.compute.astro
```
rhiannonlynne, line 10: I think this has been bothering me (and maybe other things in this same part of schedview) because I don't know if I understand the use case. I can see that "things are calculated with MAF" here, but in general, wouldn't you want to pre-calculate the MAF metrics and then only show the outputs here, so that any "calculation of things with MAF" should actually happen with a batch and script that calls MAF directly? Rather than running it via schedview? 

And if you had a bundle that you wanted to start execution of with MAF, wouldn't the better approach be to set up a bundle and script within MAF, and then execute that script (in rubin_sim) with a trigger from schedview?

rhiannonlynne (approved):
I have some questions, a few typos, and a comment about the placement of __all__ that should be addressed.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: check-toml
+  - repo: https://github.com/psf/black
```
timj, line 9: FYI, black now has a special pre-commit implementation so you can change the URL to https://github.com/psf/black-pre-commit-mirror (but need to uprev the black version first with `pre-commit autoupdate`).

pyproject.toml
```diff
@@ -31,3 +31,51 @@ __version__ = "{version}"
...
+line-length = 110
+select = [
+    "E",  # pycodestyle
+    "F",  # pyflakes
+    "N",  # pep8-naming
+    "W",  # pycodestyle
+    #"D",  # pydocstyle
+]
+target-version = "py310"
```
timj, line 72: py311?

schedview/compute/__init__.py
ehneilsen: I'm moving the `__all__` assignments to the top. Although tuples seem to work in practice, the [official python.org documentation](https://docs.python.org/3/tutorial/modules.html), as well as the examples in PEP8, all say that `__all__` should be a list, so I think we should leave them as lists.

schedview/compute/maf_metrics.py
ehneilsen: If I recall correctly, I wrote this for use in notebooks, without dashboards in mind. In any case, this file isn't used anywhere in the dashboards, and I'm not using those notebooks anymore, so the file can just be removed, I'll do that.

ehneilsen: I do think there will need to be a home for utilities that compute stuff with maf that are needed for specific plots needed on rare occasions (or only for parameters that are bespoke to a few situations), but not generally applicable enough to be computed routinely or automatically. I don't have an opinion on whether this code should be part of maf or schedview, though.

PR #20, opened 2023-09-21
Remove example user-defined plots from default

ehneilsen:


rhiannonlynne (approved):
I noticed this came up in the default plots in the app and was curious why.

PR #24, opened 2023-09-26
tickets/PREOPS-3769: Expand schedview documentation

ehneilsen:
Assorted expansions and improvements to the docs directory (and corresponding content on schedview.lsst.io).
See https://schedview.lsst.io/v/PREOPS-3769/index.html


rhiannonlynne:
Few minor typos in the docs (check 'simulation' mostly I think) but nothing major.
I was a bit confused by the architecture page though, because it's giving examples of things which are not implemented in schedview and also are not things we would want to show in schedview. Perhaps some of this confusion could be alleviated by making a specific statement at the start along the lines of "To demonstrate the principles underlying the chosen architecture in `schedview`, this walk-through uses an example currently outside the scope of our dashboards. This example demonstrates using the schedview framework to create a dashboard showing the location of minor planets in the sky on a given date."  (unless I've missed something entirely).

docs/introduction.rst
```diff
@@ -1,11 +1,43 @@
...
-- `sched_maps`, a tool for examing some elements of the state of objects used
-  by the scheduler, particularly those that take the form of a sky map,
-  although some indications of other elements are also present.
-- `prenight`, a pre-night briefing dashboard.
+``schedview`` presently supplies two dashboards:
+
+* ``scheduler_dashboard``, a dashboard for examining the state of a the scheduler an a specific time, intended for using during the night to give insight into the schedulers current behavior.
+  The ``scheduler_dashboard`` shows maps and relevant statistics for all basis functions and rewards for all surveys, given a time and an instance of the scheduler.
+* ``prenight``, a dashboard for preparing for a night of observing.
+  Visualizations included in this dashboard are intended to help schedulers and observing staff understand what the scheduler will to in a night, given the output of a simualtion of the night (as created by `rubin_sim`).
```
rhiannonlynne, line 16: ".. what the scheduler will *do* in a night, given the output of a *simulation* of .. "

rhiannonlynne (approved):

ehneilsen:
Yes, that was exactly my intention with the architecture page. The example was chosen to be as simple as possible while including all of the architectural elements, and using real data.

I did consider using completely artificial data (loading a table with columns 'foo' and 'bar', which get put through some random equation before being plotted as a scatter plot). Tomorrow, I can move to doing that if you think it will be less misleading, or I can add text like you've proposed.

I did try to emphasize in the text that everything in schedview is intended to be specific to the scheduler or progress tracking.

rhiannonlynne:
Where did something like getting the scheduler pickle (for collect) // computing a single current basis function // plotting it // making the dashboard (or a much simplified single plot dashboard?) 
break down? Is there a way to find a simple example using some of the data we'd actually use? 
Alternatively, I see you had a previous version without code examples .. did you feel like it didn't get the point across well enough? 
I'm ok with leaving it like this, especially with the disclaimer about why all of a sudden we're talking about minor planets. 
I'd just like to understand if the alternatives were just too cumbersome and if there's a way to simplify things to use examples that are more related to what we're actually doing. (presumably the point here is to explain why things are set up the way they are, for future developers to be able to follow the same design choices).

PR #26, opened 2023-10-02
tickets/PREOPS-3790: Add option use a file selector to specify data in schedview's prenight app

ehneilsen:


rhiannonlynne:
I'm pretty sure that we're not going to be able to run dashboards that access the host filesystem (i.e. /sdf) - I wish I could remember exactly where I read it, but I saw a doc with requirements for running RSP deployed things that indicated this would not be allowed. That said, I'm not sure we have a better alternative at the moment, so I'm not going to say you have to change it .. just a heads up.

rhiannonlynne:
However, the other thing I would like you to change, please. Can you think of a better flag than `--insecure` ? While looking through the changes, until I got to the docstring of the flag, I was thinking that --insecure meant "this is operating in an insecure environment, so lock down the html upload option" .. the flag indicates that it's actually the other way around. So, basically, maybe it would be good to find something that is less ambiguous? 
--allow_uploads or something? or if the default IS to allow uploads, then --no_uploads? or something better.

rhiannonlynne (changes requested):
Please find a non-ambiguous flag name.

ehneilsen:
> I'm pretty sure that we're not going to be able to run dashboards that access the host filesystem (i.e. /sdf) 

What do they mean by "access"? Obviously they don't mean it in the most general sense, because that would be nearly equivalent to a blanket ban on dashboards. After all, the dashboard code relies on the operating system, which generally requires access to the host filesystem.

I have two speculations about what they might actually mean. The first is that they might mean that we can't allow downloads of files based on directory listings. This is a pretty common policy for web servers. But, schedview does not allow download of any data files; all it does is allow users to make plots from them. So, I don't think schedview would break this policy.

The other thing they might mean is that the dashboard must run in a container that does not mount any host filesystems. This isn't quite the same as not having access to the host filesystem; after all, everything in the container is still on the host filesystem. On the other hand, the programs in the container cannot see the host filesystem as a filesystem. This would be an irritation for schedview, but not a show-stopper. We would just need to copy all of the data available for viewing from the host filesystem to the container filesystem when it gets started, and restart the container whenever new data gets added. This would make the containers big and slow to start, but for what we want to do with them right now, I think this is probably fine. After all, this is not meant to be a permanent solution, just something we can use for debugging and demonstration until proper data archive (maybe butler based) is implemented.

ehneilsen:
> Please find a non-ambiguous flag name.

I just pushed a change that renames the option from --insecure to --data_from_urls, which explicitly states what it does.

I would be happier if I could think of a short and unambiguous flag name that is more alarming. The security concern for why you would not want to use this flag is not obvious, and I want to be sure nobody uses this option without being alerted to the danger.

rhiannonlynne:
Maybe have the program check the flag and report a warning if it's set to allow uploads? This doesn't help with the automatic dashboard setup where users may not see the log, but in that situation we should have tighter control over what's happening anyway.

ehneilsen:
I added a confirmation dialog, together with a --no_confirmation flag that prevents it from showing the confirmation dialog (for starting automatically but with restricted access, which I think might be what is wanted at the observatory).

rhiannonlynne:
> 
> The other thing they might mean is that the dashboard must run in a container that does not mount any host filesystems. This isn't quite the same as not having access to the host filesystem; after all, everything in the container is still on the host filesystem. On the other hand, the programs in the container cannot see the host filesystem as a filesystem. This would be an irritation for schedview, but not a show-stopper. We would just need to copy all of the data available for viewing from the host filesystem to the container filesystem when it gets started, and restart the container whenever new data gets added. This would make the containers big and slow to start, but for what we want to do with them right now, I think this is probably fine. After all, this is not meant to be a permanent solution, just something we can use for debugging and demonstration until proper data archive (maybe butler based) is implemented.


We talked about this in person, but for later .. yeah, it's that the container is not supposed to use host file paths. 
Peter has suggested checking the size of the skybrightness files for various date ranges, and then balancing that against what would be reasonable range to expect to have accessible in the container at any time .. for example, a year's worth may not be too slow/large and would probably contain more than enough skybrightness information at any given time, but I don't know what the limits or sizes are.

rhiannonlynne (approved):

rhiannonlynne (approved):

PR #32, opened 2023-10-06
tickets/PREOPS-3843: add --show option to prenight

ehneilsen:
Adds a --show option to `prenight` which causes it to show the dashboard in a new browser window.

### DIFF
--- a/schedview/app/prenight/prenight.py
+++ b/schedview/app/prenight/prenight.py
@@ -1026,6 +1026,12 @@ def parse_prenight_args():
     )
 
     parser.add_argument(
+        "--show",
+        action="store_true",
+        help="Open a browser window to show the dashboard.",
+    )
+
+    parser.add_argument(
         "--port",
         "-p",
         type=int,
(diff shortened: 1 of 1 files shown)


schedview/app/prenight/prenight.py
```diff
@@ -1065,6 +1071,12 @@ def main():
 
     del prenight_app_parameters["yes"]
 
+    try:
+        show = prenight_app_parameters["show"]
+        del prenight_app_parameters["show"]
+    except KeyError:
+        show = False
```
rhiannonlynne, line 1078: So .. I don't think you need this try/except, unless I'm missing something here. 
Just add a default of "False" to your add_argument definition up above, and then it will always be False unless the user sets it to be True. 
I'm also not entirely clear why you need to drop the extra keys in the prenight_app_parameters dictionary, but didn't really look in depth at what you're doing with it.

rhiannonlynne: Ok .. I went and took a look. I seems like it's so that you can keep a boundary between parameters for a piece of code that actually runs the prenight app vs. the dashboard (presumably to keep a separation between how the dashboard is configured / what port /etc. and the app itself).

rhiannonlynne (approved):

PR #36, opened 2023-10-09
Allow prenight data files to be set on command line even if arbitrary…

ehneilsen:
… URLs are not allowed

### DIFF
--- a/schedview/app/prenight/prenight.py
+++ b/schedview/app/prenight/prenight.py
@@ -5,6 +5,7 @@ import logging
 import os
 import sys
 from glob import glob
+from pathlib import Path
 
 import astropy.utils.iers
 import bokeh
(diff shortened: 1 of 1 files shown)


schedview/app/prenight/prenight.py
```diff
@@ -932,7 +947,16 @@ def prenight_app(*args, **kwargs):
...
+        specified_data_files = {}
+        data_args = set(["opsim_db", "scheduler", "rewards"]) & set(kwargs.keys())
+        for data_arg in data_args:
+            # Fully resolve the path to the file.
+            file_path = Path(kwargs[data_arg]).resolve()
+            kwargs[data_arg] = str(file_path)
+            if data_arg in kwargs:
+                specified_data_files[data_arg] = str(file_path)
+
+        prenight = RestrictedInputPrenight(data_dir=data_dir, **specified_data_files)
```
rhiannonlynne, line 959: Oh interesting .. so I think this means that if you do not allow arbitrary data urls (which I thought meant uploading files only, rather than files from the filesystem), that you have to pre-specify which data directories should be allowed? 
So the choices are -- data files from paths that are specified at startup vs. any data file + URL uploads? 
(I'm not sure if that would be clear to other people, but wasn't clear to me, if that's how it works)

rhiannonlynne (approved):
Might need to clarify some documentation, depending on if my interpretation of why these changes were necessary/what they allow is correct.

schedview/app/prenight/prenight.py
ehneilsen: Yes, that's right.

PR #37, opened 2023-10-12
accept nside, scheduler file name and mjd as url parameters

emanehab99:
This PR allows the scheduler dashboard to accept scheduler file name, nside and date mjd value as url paramaters.
- use `scheduler` to set scheduler file name (path or URL)
- use `nside` to set nside value
- use `mjd` to set date/time in mjd format

Example:
`http://localhost:8080?nside=16&scheduler=%2Fhome%2Feman%2FDownloads%2Fsched_maps%2Fexample_pickle_scheduler.p.xz&mjd=60207`


ehneilsen (approved):
Fix the tests so they pass in github before merging, but otherwise looks good.

PR #38, opened 2023-10-13
PREOPS-3752: Add survey fields and telescope pointing to map.

alserene:
- Pass newly added conditions and survey parameters to schedview.plot.survey.map_survey_healpix()

- Update Bokeh plot of key to match colour and alpha values of plot objects.

### DIFF
--- a/schedview/app/scheduler_dashboard/scheduler_dashboard.py
+++ b/schedview/app/scheduler_dashboard/scheduler_dashboard.py
@@ -761,6 +761,8 @@ class Scheduler(param.Parameterized):
                 self._survey_maps,
                 "above_horizon",
                 self.nside,
+                conditions=self._conditions,
+                survey=self._scheduler.survey_lists[int(self._tier[-1])][self._survey],
             )
             self._sky_map_base.plot.toolbar.tools[-1].tooltips.remove(("above_horizon", "@above_horizon"))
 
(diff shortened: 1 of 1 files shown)


ehneilsen (approved):
Looks good!

Eventually (but not needed for this issue) we may want to change how the legend is made so that the colors etc. are automatically kept the same, and we do not rely on hard-coding the values in two separate places. I can think of several different approaches:

1. Use the legend-creating tools built into bokeh
2. Leave the legend creation separate (as it is), but have the colors etc. read from a common location.
3. Leave the legend creation separate, but instead of hard-coding the colors there, query the bokeh plot object for the colors used.

As a general rule, whenever two values are hard-coded separately but need to be the same, it makes maintaining the code long term harder.

alserene:
> Looks good!
> 
> Eventually (but not needed for this issue) we may want to change how the legend is made so that the colors etc. are automatically kept the same, and we do not rely on hard-coding the values in two separate places. I can think of several different approaches:
> 
> 1. Use the legend-creating tools built into bokeh
> 2. Leave the legend creation separate (as it is), but have the colors etc. read from a common location.
> 3. Leave the legend creation separate, but instead of hard-coding the colors there, query the bokeh plot object for the colors used.
> 
> As a general rule, whenever two values are hard-coded separately but need to be the same, it makes maintaining the code long term harder.

I agree that this is something we should do 👍 I'll have a think on the options.

PR #39, opened 2023-10-18
Utility to download pickles from the EFD

ehneilsen:
This code is just to help developers when developing at the USDF. Corresponding code to use the EFD from dashboards is covered by a separate issue, PREOPS-4097. Addressing PREOPS-4097 faces several obstacles (authentication, integration with dashboard code) which need not be blockers for the use cases addressed by PREOPS-4096, so PREOPS-4096 is separated out.


timj (commented):
As a general comment is there a plan to stop using pickle? It's a terrible archive format if you want these pickle files to be readable in the future.

ehneilsen:
> As a general comment is there a plan to stop using pickle? It's a terrible archive format if you want these pickle files to be readable in the future.

I completely agree that it's a terrible archive format (and that's not its only problem), and there are chronic discussions within the scheduler team of adopting a format better suited to long term archiving. Several alternatives have been discussed, and when it comes to long term archiving, we may actually end up doing one of them. But, the scheduler is complex enough that it isn't something that can be done without real thought and work.

For our most pressing use case, though, pickles have been good enough, because it isn't about long-term archiving, but reading it only a short period of time after it was made (minutes to months, not years).

rhiannonlynne:
Indeed, as Eric said  - the pickles are intended to be "snapshots" of the scheduler as in-use on the telescope (so created and retrieved within the night or a few days), and longer-term retrieval or use should be able to start with the scheduler (python) configuration at the start of the night and then replay observations onto it, using "Conditions" saved at the telescope .. the conditions are currently pickles and need to change, but this is at least simpler than changing the scheduler object (as Conditions is essentially a dictionary).

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+        Otherwise, the file path of the saved scheduler is returned.
+
+    Notes
+    -----
+    This function sets the environment variable
+    `LSST_DISABLE_BUCKET_VALIDATION` to "1".
+    """
+
+    os.environ["LSST_DISABLE_BUCKET_VALIDATION"] = "1"
+    scheduler_resource_path = ResourcePath(scheduler_url.replace("https://s3.cp.lsst.org/", "s3://rubin:"))
```
rhiannonlynne, line 111: I think this access to the LFA needs to be location dependent also. From the summit, you do actually use s3.cp.lsst.org and not s3://rubin. 

Perhaps this argues for setting "location" for both get_scheduler and the methods above, which would be either of "summit" or "usdf"? (summit EFD = "summit_efd", usdf EFD = "usdf_efd", summit LFA = "https://s3.cp.lsst.org/<path>" usdf LFA = "s3://rubin:<path>")  That way it's the same kwarg for all of these methods, and we can just default to usdf (but then a user could just substitute "summit" in all places if at summit). 
In addition, I think it would likely future-proof this a bit if instead of replacing the summit URL explicitly, we split the path (ResourcePath.path() will return the path .. might have to rstrip('/') still, but then add it onto the rest of the url?)

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+    Parameters
+    ----------
+    desired_time : `str`
+        The central time to query the event service in ISO format
+        (YYYY-MM-DDTHH:MM:SS).
+    efd : `str`, optional
+        The name of the EFD to connect to (default is 'usdf_efd').
+    time_window_seconds : `float`, optional
+        The size of the time window to search for scheduler URLs, in seconds
+        (default is 2).
```
rhiannonlynne, line 30: Does it make sense to make "time" and "time_window" be astropy Time and DeltaTimes? Or is it not working so well with the place you want to place this within the dashboard, because the dashboard wants TimeStamps or strings?

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+        A DataFrame containing the scheduler snapshot times URLs.
+    """
+    desired_time = Time(desired_time)
+    start_time = desired_time - (time_window_seconds / 2) * u.second
+    end_time = desired_time + (time_window_seconds / 2) * u.second
+
+    efd_client = EfdClient(efd)
+    topic = "lsst.sal.Scheduler.logevent_largeFileObjectAvailable"
+    fields = ["url"]
+    scheduler_urls = await efd_client.select_time_series(topic, fields, start_time, end_time)
```
rhiannonlynne, line 44: It likely doesn't matter at the moment, but I believe the salIndex = 2 means "for aux tel" while 1 would be main tel. 
Without definition, I don't know what you get (at the moment only auxtel is active, so we just get auxtel).
Probably we just need to add a kwarg so that we can specify which to retrieve, as the time is coming not too far in the future where we'd want main tel things too.

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+            raise ValueError("night must be a valid date in YYYY-MM-DD format")
+
+    # The offset by longitude moves to local solar time.
+    # The offset of 1 makes the night specificed refer to the
+    # local date corresponding to sunset.
+    local_midnight = Time(reference_time.mjd + 1 - Site("LSST").longitude / 360, format="mjd")
+    time_window_seconds = u.day.to(u.second)
+    scheduler_urls = await query_schedulers_in_window(
+        local_midnight, efd=efd, time_window_seconds=time_window_seconds
+    )
```
rhiannonlynne, line 82: same thing about the sal index

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+
+async def get_scheduler_on_night(night, scheduler_index, destination, efd="usdf_efd"):
+    """Get a scheduler for a given night from the EFD.
+
+    Parameters
+    ----------
+    night : `str`
+        The night for which to get the scheduler.
+    scheduler_index : `int`, optional
+        The index of the scheduler to retrieve (default is 0).
```
rhiannonlynne, line 171: I suspect that "index" is an overloaded term, when there is also salIndex running around. 
In general, I'd suggest this does not need to exist as-is .. people can get the list of all the schedulers in a night above, then pick out the one they want to load, and then load it. 
This being a simpler thing of "get the latest scheduler" would make sense though. 

Note that "these_scheduler_references.url[index]" is triggering warnings in the most recent version of pandas and probably needs to swap to these_scheduler_references.url.iloc[index].

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+async def main():
+    parser = argparse.ArgumentParser(description="Query the EFD for a scheduler.")
+    parser.add_argument(
+        "datetime", help="The desired time or night to get the scheduler for in iso8601 format."
+    )
+    parser.add_argument(
+        "--destination", type=str, default=None, help="The file path to save the scheduler to."
+    )
+    parser.add_argument(
+        "--id", type=int, default=0, help="The index of the scheduler to retrieve (default is 0)."
```
rhiannonlynne, line 198: Yeah -- I read this as "index = 1 or 2" for "main tel or aux tel" .. not "what number of the scheduler snapshots within the night should I load".

util/efd_sched.py
```diff
@@ -0,0 +1,231 @@
...
+        The scheduler for the given night.
+    """
+    these_scheduler_references = await query_night_schedulers(night, efd=efd)
+    this_url = these_scheduler_references.url[scheduler_index]
+    result = get_scheduler(this_url, destination)
+    return result
+
+
+async def main():
+    parser = argparse.ArgumentParser(description="Query the EFD for a scheduler.")
```
rhiannonlynne, line 190: I don't think I understand this script .. does it returns the scheduler file name if you specify "destination" but simply returns a list of scheduler files if you do not? This seems like a mix of two things which are actually separate .. does it even need to be a "main" function?

rhiannonlynne (changes requested):

util/efd_sched.py
ehneilsen: This is a bit of an expansion of scope for this issue. I had not intended for this code ever to be run anywhere other than the USDF: it's a development tool, not a production too. It is unlikely ever to be used once PREOPS-4097 is addressed, which I expect to be relatively soon, so "future-proofing" it seems a bit pointless. Still, this seems straightforward, so I'll go ahead and do it.

util/efd_sched.py
ehneilsen: I don't expect this code to live long enough for anything other that auxtel to be relevant. There's no harm in supporting the main telescope as well, though.

util/efd_sched.py
ehneilsen: I'm mostly trying to avoid doing extra typing or cutting-and-pasting by letting the user pick a small integer instead. How about if I replace the word `index` with `row_number`?

util/efd_sched.py
ehneilsen: Again, I'll replace `id` with `row_number`.

util/efd_sched.py
ehneilsen: It can do either. I'll rework the command line to make it clearer.

util/efd_sched.py
ehneilsen: Done.

util/efd_sched.py
ehneilsen: Done.

util/efd_sched.py
ehneilsen: Done.

util/efd_sched.py
ehneilsen: Done.

rhiannonlynne (approved):

PR #41, opened 2023-11-01
tickets/PREOPS-4125: update for param 2.0.0

ehneilsen:
With only a handful of pretty minor changes, all `pytest` tests pass, the notebooks run, and walking through the `prenight` and `scheduler_dashboard` with some sample data all work as before.
I do not _think_ any of the changes would prevent it from running with older version of `param` as well, but I have not tested this as thoroughly.


rhiannonlynne (approved):

rhiannonlynne:
I'm a little confused because the commit which says "update for param 2.0" seems like it was maybe just rerunning the notebooks and adjusting to changes in the underlying schedview api? But I think the only actual changes to the code that were required for the param update were in the first two commits?

ehneilsen:
> I'm a little confused because the commit which says "update for param 2.0" seems like it was maybe just rerunning the notebooks and adjusting to changes in the underlying schedview api? But I think the only actual changes to the code that were required for the param update were in the first two commits?

The biggest change in the notebook by number of characters was accidental (and not important): they were originally saved unexecuted, so the repo did not include the output, and in at least one case I forgot to clear the output before committing it to git.

The notebooks did have some actual code changes related to updating to param 2.0, for example the new version seems to get confused by the `./` in front of file names in the file selector, so I took them out. I think I made a couple of other minor changes (e.g. adding comments). Mostly, though, not much actually changed in the notebooks.

PR #43, opened 2023-11-02
PREOPS-4512 Serialization error when choosing nside value from the dropdown

emanehab99:
Define nside values as int list rather than np.int64 array to make th…e sync between with the nside URL parameter seamless

### DIFF
--- a/schedview/app/scheduler_dashboard/scheduler_dashboard.py
+++ b/schedview/app/scheduler_dashboard/scheduler_dashboard.py
@@ -127,7 +127,7 @@ class Scheduler(param.Parameterized):
     )
     nside = param.ObjectSelector(
         default=16,
-        objects=[2**n for n in np.arange(1, 6)],
+        objects=[2, 4, 16, 32, 64],
         label="Map resolution (nside)",
         doc="",
     )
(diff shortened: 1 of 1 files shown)


schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -127,7 +127,7 @@ class Scheduler(param.Parameterized):
     )
     nside = param.ObjectSelector(
         default=16,
-        objects=[2**n for n in np.arange(1, 6)],
+        objects=[2, 4, 16, 32, 64],
```
alserene, line 130: The dashboard currently displays nside options [2, 4, 8, 16, 32].
@ehneilsen can you clarify which options should be given?

ehneilsen: These are fine. Mostly I think people will want 16, 32, or 64, or maybe 8 if the user is impatient or testing. I doubt 2 or 4 will get used much, but there presence is harmless. At present, higher than 64 is not useful, because none of the instances of scheduler we have right now use nside>64. I suppose its possible that someday there might be data with higher nsides than this, but I don't think we should worry about adding them for something so speculative.
If you wanted to be really fancy, you could offer the powers of 2 up to and including `scheduler.nside`, where `scheduler` is whatever instance of the scheduler is loaded, but really I don't think it's worth the effort to make it dynamic like this.
`nside` is _always_ a power of 2.

ehneilsen (approved):

PR #44, opened 2023-11-02
Tickets/PREOPS-4400: Date does not update correctly.

alserene:
- Set date from the conditions object of a loaded pickle file.
- Simplify date-related functions to manage pickle, url and widget date inputs.
- Modify attribute names to remove confusion.

### DIFF
--- a/schedview/app/scheduler_dashboard/scheduler_dashboard.py
+++ b/schedview/app/scheduler_dashboard/scheduler_dashboard.py
@@ -113,8 +113,13 @@ class Scheduler(param.Parameterized):
         label="Scheduler pickle file",
         doc=scheduler_fname_doc,
     )
-    date = param.Date(default=DEFAULT_CURRENT_TIME.datetime.date(), doc="")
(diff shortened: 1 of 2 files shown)


schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -487,14 +543,57 @@ def read_scheduler(self):
...
+
+        try:
+            self._debugging_message = "Starting to update Conditions object."
+            pn.state.notifications.info("Updating Conditions object...", duration=0)
+
+            # self._conditions.mjd = self._mjd
+
+            # Use instance of ModelObservatory until Conditions
+            # setting bug is fixed.
+            self._model_observatory = ModelObservatory(
```
ehneilsen, line 567: If `self._model_observatory` already exists and has the correct nside, you should be able to just reuse it and avoid the significant time it takes to create a new instance.

ehneilsen (approved):

PR #45, opened 2023-11-02
Tickets/PREOPS-4401: Rewards table displays data for the wrong survey.

alserene:
- Index fields added to df output by make_scheduler_summary_df() to maintain integer ordering.
- Index column added to scheduler summary table.
- Additionally, survey_name field changed to avoid overloading.


ehneilsen (approved):
Looks good.

PR #46, opened 2023-11-03
Update tests; consolidate pypi build and publish

rhiannonlynne:
Reworking the workflows .. but I may need to merge this before being able to check, due to git mv for one of the workflows.


rhiannonlynne:
Oh, that's interesting. It's running both the version of "run unit tests" from main and the version from here (the workflow yaml files  for  'python_test' and 'test_and_build").

rhiannonlynne:
Wow, it is so weird how *this repo* has an issue with the ubuntu version of the download.  Working on a thing to investigate.

.github/workflows/test_and_build.yaml
```diff
@@ -27,39 +37,48 @@ jobs:
...
-          tar -xf site_models_2023_03_28.tgz
-          echo "site_models_2023_03_28.tgz complete"
-          wget -q https://s3df.slac.stanford.edu/data/rubin/sim-data/rubin_sim_data/scheduler_2023_09_22.tgz
-          tar -xf scheduler_2023_09_22.tgz
-          echo "scheduler_2023_09_22.tgz complete"
-          wget -q https://s3df.slac.stanford.edu/data/rubin/sim-data/rubin_sim_data/skybrightness_pre_2023_09_19.tgz
-          tar -xf skybrightness_pre_2023_09_19.tgz
-          echo "skybrightness_pre_2023_09_19.tgz complete"
+          export RUBIN_SIM_DATA_DIR=~/rubin_sim_data       
+          rs_download_data --force --dirs scheduler,site_models,skybrightness_pre,throughputs,tests --tdqm_disable
```
ehneilsen, line 75: Does this actually work here?

ehneilsen (approved):
Looks good, as far as I understand it.

.github/workflows/test_and_build.yaml
rhiannonlynne: Well ... I haven't fallen into the problem of needing to use it. 
It did work from the cache download script. 
This should only trigger if the version of rubin_sim changed since the cache was created. The cache will be re-created every night, or can be manually triggered. 
So it's just a backup, really. We could copy in the full "download all these files individually", but it seems like not a great option. It could also be deleted entirely, which was my first thought .. and then I figured, no it was worth leaving in as a "good example". 
It's also worth noting that the cache build script is currently running on an ubuntu runner, but it would be perfectly fine to move that to a macOS runner if needed.

rhiannonlynne: If you wanted me to pull the download individual files option into a different file, to preserve for easy pasting later, I could do that?

PR #47, opened 2023-11-07
PREOPS-4518: Preserve wind and seeing conditions from a loaded pickle.

alserene:


schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -568,14 +569,34 @@ def update_conditions(self):
...
+                seeing_data = rubin_sim.site_models.ConstantSeeingData(0.69)
+                # Create new MO instance.
                 self._model_observatory = ModelObservatory(
                     nside=self._scheduler.nside,
                     init_load_length=1,
+                    wind_data=wind_data,
+                    seeing_data=seeing_data,
                 )
 
+            # Code for tickets/PREOPS-4817
```
ehneilsen, line 587: In general, if code has previously been checked into git, just deleting it rather than commenting it is usually a better approach, although a text comment describing why one approach was taken rather than the other, maybe mentioning the commit where the other approach was taken, can be quite useful. Usually, "why" comments are more useful for posterity than "what" comments. (There are, of course, exceptions to this rule of thumb.)

(Just because I know this is good advice doesn't mean I actually always follow it myself, but I should.)

ehneilsen (approved):
Provided that `update_conditions` does not get called when a new pickle is loaded, this looks fine. (After a pickle is loaded but before the user sets a new time, the conditions instance that should be being used is the one from the pickle, not one generated by the dashboard's instance of `ModelObservatory`. I _think_ that's what you've done, but I'm not 100% sure. One challenge with `param` is that it's sometimes hard to tell what triggers what.)

schedview/app/scheduler_dashboard/scheduler_dashboard.py
alserene: Yes, `update_conditions` only gets called when the date is changed 👍

The commented code is actually for the next ticket related to this, [ticket 3817](https://jira.lsstcorp.org/browse/PREOPS-3847), allowing users to set weather conditions. (I realise now I've written the ticket number wrong.) I'll just remove the commented code and store in a file on my computer somewhere instead.

emanehab99:
Looks good! Thanks for updating the tests

PR #50, opened 2023-11-16
Tickets/preops 3846

emanehab99:
added a subsclass of Scheduler that uses param FileSelector to restrict files to a certain directory, added a cli argument `--data_dir `to set data directory and choose between normal dahsboard behaviour or restricted one. URL paramaters don't work in restricted mode

add `--data_from_urls` argument to switch to non-restricted mode

Restricted mode is the default. If --data_dir isn't provided, files will be restricted what is in the `data` directory in the code base

### DIFF
--- a/schedview/app/scheduler_dashboard/scheduler_dashboard.py
+++ b/schedview/app/scheduler_dashboard/scheduler_dashboard.py
@@ -35,7 +35,6 @@ from datetime import datetime
 from glob import glob
 from zoneinfo import ZoneInfo
 
-import astropy.utils.iers
 import bokeh
 import numpy as np
 import panel as pn
@@ -59,6 +58,7 @@ import schedview.compute.survey
 import schedview.param
 import schedview.plot.survey
 
+# filter astropy warning that's filling the terminal with every update
 warnings.filterwarnings("ignore", category=AstropyWarning)
 
 DEFAULT_CURRENT_TIME = Time.now()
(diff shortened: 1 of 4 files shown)


emanehab99:
@ehneilsen I cannot understand the unit test failures. It says maximum regression depth reached when initialising the restricted file selector. Not sure what that is and I cannot replicate it locally (I get other KeyErrors related to rubin sim!)

emanehab99:
I've created a new environment with updated rubin sim. So now I get the same error locally. But still cannot understand the reason

emanehab99:
@ehneilsen This version of the dashboard with the restricted input is working perfectly with param v1. Now that the codebase is using param v2, this version doesn't work. I've created another version that works with param v2, however, pickles are always loaded from PACKAGAE_DATA_DIR regardless of any other path provided in the commandline arguments.
I've tried the prenight dashboard and it seems like it behaves the same. Can you please confirm that?

schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -1274,6 +1285,32 @@ def map_title(self):
...
+        doc=scheduler_fname_doc,
+        default="",
+        allow_None=True,
+    )
+
+    def __init__(self, data_dir=None):
+        super().__init__()
+
+        if data_dir is not None:
+            self.param.scheduler_fname.path = f"{data_dir}/*scheduler*.p*"
```
ehneilsen, line 1311: I think you should be able to fix the problem of ignoring the data_dir as set by the command line argument by replacing this assignment with something like:
`self.param['scheduler_fname'].update(path=f"{data_dir}/*scheduler*.p*")`

ehneilsen (approved):

schedview/app/scheduler_dashboard/scheduler_dashboard.py
emanehab99: Thank you. That worked

PR #51, opened 2023-11-21
tickets/PREOPS-4565 Modify schedview-prenight dashboard startup to support running within k8

ehneilsen:
This is at the point where it starts, and can look at the data file contained within the package.


rhiannonlynne:
Presumably the 'print' things would go away once debugging is done, or are they necessary for use? (or are they useful to leave there to allow for debugging within k8s later?)
.. maybe I'm just saying, some comments in the code as to the purpose might be useful for later.

rhiannonlynne (approved):

PR #52, opened 2023-11-21
tickets/PREOPS-4597: color by note in schedview prenight az/alt plt

ehneilsen:
Erik gave me a push on this this morning, so I pushed it to the top of my to do list.


rhiannonlynne (approved):
Looks fine. I could only test it with the sample_opsim.db outputs, not outputs from an auxtel setup.

rhiannonlynne:
sorry about the force push .. while I had your branch checked out, I accidentally pushed a commit here instead of to my branch-in-progress. So the force push was just removing that commit.

PR #54, opened 2023-11-27
tickets/PREOPS-4603: prepare schedview's scheduler_dashboard app for running at USDF w/ k8

ehneilsen:
The k8 ingress seems to have trouble getting supplying bokeh js libraries unless the dashboard is served a couple of directories "down" in the url, that is https://myhost/a/b works, but https://myhost and https://myhost/a do not.
There might be a way either to get bokeh to serve its libraries differently or get the ingress to handle routing to the js libraries, but I couldn't find it, and just moving the URL is easy and has no downsides that I can see.
I named the dir "scheduler-snapshot" here to make it clear how it is different from other existing planned or existing schedview dashboards, which can also be thought of as scheduler-related dashboards. What makes this dashboard distinctive is that it looks at the state of the scheduler as captured (in a pickle, at least for now) at a given time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,10 @@ RUN micromamba install -y -n base -f /home/${MAMBA_USER}/schedview/container_env
     micromamba clean --all --yes
 ARG MAMBA_DOCKERFILE_ACTIVATE=1
 RUN python -m pip install /home/$MAMBA_USER/schedview --no-deps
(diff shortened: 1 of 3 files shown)


alserene (approved):
As far as I can tell, this all looks okay. I ran the dashboard locally in both the restricted and unrestricted modes and they both worked fine.

PR #58, opened 2023-12-08
tickets/PREOPS-4647: update night in prenight when visits loaded, if needed

ehneilsen:


rhiannonlynne:
Yeah, the codecov report seems odd in other repos as well. I'm not entirely sure what to do about it. 
The report seems not necessarily in line with what seems reasonable to me either. 
I suppose a minimum thing to do would be to take the badges off the front pages (of our other repos, not added yet for schedview).

rhiannonlynne (approved):
Does this mean the dashboard will start up on the middle night out of a set of visits? I still think the first night is probably where you'd like to start, but I suppose further testing will help understand better what's most useful.

ehneilsen:
Yes, currently it starts in the middle. I chose that rather than the first because I've seen sims that just have one or two exposures on one night, followed by a full night on the next, in which case it's clearly the next night that is wanted.

A better approach might be use the date of the first night with more than some number of exposures, but really, when faced with a database with multiple nights, anything the code might do is a just a wild guess anyway.

I can imagine that in the future that the code that automatically generates sims for the pre-night briefing will do exactly some number of nights, and reliably _not_ start in the middle of a night, in which case it will either not matter (if it does one), or will be easy to change to do the first night.

PR #59, opened 2023-12-12
tickets/PREOPS-4685: use lsst.resources for loading data in the prenight briefing

ehneilsen:
Right now, it's only been tested on files in the local filesystem and package resources, but hopefully it will work with s3 buckets as well, when there is one with opsim data.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dependencies = [
     "param",
     "pytz",
     "rubin-scheduler",
+    "lsst.resounces",
     "uranography >= 1.1.0 ",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "param",
     "pytz",
     "rubin-scheduler",
-    "lsst.resounces",
+    "lsst.resources",
     "uranography >= 1.1.0 ",
 ]
 
--- a/schedview/app/prenight/prenight.py
+++ b/schedview/app/prenight/prenight.py
@@ -1,10 +1,9 @@
 import argparse
-import importlib.resources
 import json
 import logging
 import os
 import sys
-from glob import glob
+import urllib.parse
 from pathlib import Path
 
 import astropy.utils.iers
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "param",
     "pytz",
     "rubin-scheduler",
+    "lsst.resources",
```
rhiannonlynne, line 37: I think you're right that lsst.resources will install from pypi, but I'm guessing that for conda-forge it will have to be lsst-resources (and I'd like to ask for a conda-forge package). 
And also, lsst-resources *does* work for pypi as well, so maybe it'd be good to just simply use "lsst-resources" instead of lsst.resources in one place and lsst-resources in another? 
(and I also agree that for pyproject.toml it makes no difference, given how pip checks .. it's entirely for looking 'the same').

schedview/app/prenight/prenight.py
```diff
@@ -928,13 +943,12 @@ def parse_prenight_args():
...
-    default_data_dir = f"{USDF_DATA_DIR}/*" if os.path.exists(USDF_DATA_DIR) else PACKAGE_DATA_DIR
     parser.add_argument(
-        "--data_dir",
+        "--resource_uri",
         "-d",
         type=str,
-        default=default_data_dir,
-        help="The base directory for data files.",
+        default=DEFAULT_RESOURCE_URI,
+        help="The base URI for data files.",
```
rhiannonlynne, line 951: Can this still be just a directory on disk? (if so, might be helpful to say something like that .. "The base URI for data file. This could be the S3 bucket or a directory on disk." ??

rhiannonlynne (approved):
So .. I didn't actually try it (I'm going to assume the functionality is something you've tested and that it works). Seems ok to me otherwise.  I don't have an S3 bucket to attempt to retrieve from, nor things on disk (or rather, I do have stuff on disk, but they're all in the expected location, so it's harder to know if the "test" succeeded or if it just defaulted to something that worked).

schedview/app/prenight/prenight.py
ehneilsen: It will definitely take a string like `file://home/someuser/data/foo`. That's one of the ways I've been testing it. I think it may also take `/home/someuser/data/foo`, but I'm not sure.

PR #60, opened 2024-01-05
tickects/PREOPS-4646: Support addition of opsim data to an archive for use by schedview-prenight

ehneilsen:
This update to schedview needs to be coordinate with the corresponding branch (with the same name) on rubin_scheduler.


ehneilsen:
Of course, the tests in the github actions are filing now, because it isn't grabbing the rubin_scheduler version on the needed branch.

schedview/collect/opsim.py
```diff
@@ -28,10 +30,29 @@ def read_opsim(opsim_uri, start_time="2000-01-01", end_time="2100-01-01"):
...
+        sim_metadata = yaml.safe_load(metadata_path.read().decode("utf-8"))
+        obs_basename = sim_metadata["files"]["observations"]["name"]
+        obs_path = original_resource_path.join(obs_basename)
+    else:
+        # otherwise, assume we were given the path to the observations file.
+        obs_path = original_resource_path
+
+    # ResourcePath.as_local runs into threading problems when used with
+    # bokeh/panel, so write the file to a temporary directory and read it
+    # "by hand" here.
```
rhiannonlynne, line 47: Huh -- so we can't read directly from the S3 bucket into the dashboard, we have to grab it, write it to a local disk (such as in the dashboard container), and then can load it into the dashboard? 
I may be misinterpreting this comment plus the code below ..

schedview/data/sim_metadata.yaml
```diff
@@ -0,0 +1,5 @@
+files:
+  observations:
+    name: 'sample_opsim.db'
+  rewards:
+    name: 'sample_rewards.h5'
```
rhiannonlynne, line 5: Is this what a sim_metadata file for multiple simulations would look like? I don't see the date of the simulation or any of the extra metadata here, so I'm struggling a bit to understand what the sim archive metadata file really looks like, and how it's used with the updated code in this PR, in particular the parts where directories are searched and the required files are pulled in.

schedview/collect/opsim.py
rhiannonlynne: Is this because the rest of the dashboard needs to have an actual file to read?

rhiannonlynne (commented):
Just have some questions.

docs/installation.rst
```diff
@@ -63,3 +63,27 @@ Building the documentation requires the installation of ``documenteer[guide]``:
...
+
+If the section with the ``aws_access_key_id`` and
+``aws_secret_access_key`` with access to this endpoint and bucket is
+``prenight``, then the following environment variables need to be set
+in the process running the dashboard:
+
+::
+
+     $ export S3_ENDPOINT_URL='https://s3dfrgw.slac.stanford.edu/'
+     $ export AWS_PROFILE=prenight
```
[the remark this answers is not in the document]
ehneilsen: I'd be a little surprised if this was a security issue -- there credentials themselves are not here, just how to set the pointer to them.

I have changed the wording to be clear the specific example values are "for rubin use only," and that anyone needing access needs to coordinate with both the survey scheduling team and the USDF admins. Note that a non-Rubin user could set up their own bucket elsewhere, put data there, and configure schedview to use that. I'd be astonished if anyone actually did this, but there's no reason it shouldn't work fine.

schedview/app/prenight/prenight.py
```diff
@@ -876,6 +909,30 @@ def __init__(self, resource_uri=DEFAULT_RESOURCE_URI, **kwargs):
             fname_params[arg_name].objects = matching_resources
 
 
+class ArchiveInputPrenight(Prenight):
+    """A pre-night dashboard that restricts the data to files in a dir."""
```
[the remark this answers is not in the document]
ehneilsen: Of course.

schedview/app/prenight/prenight.py
```diff
@@ -885,8 +942,17 @@ def prenight_app(*args, **kwargs):
...
+        data_from_archive = kwargs["data_from_archive"]
+        del kwargs["data_from_archive"]
+    except KeyError:
+        data_from_archive = False
+
     if data_from_urls:
         prenight = Prenight()
+    elif data_from_archive:
+        resource_uri = kwargs["resource_uri"] if "resource_uri" in kwargs else None
+        prenight = ArchiveInputPrenight(resource_uri=resource_uri)
```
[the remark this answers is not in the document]
ehneilsen: You can, in fact, load data from an arbitrary external URL if you use the data_from_urls option. This is what makes this approach a security issue if it's run on shared infrastructure.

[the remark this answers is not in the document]
ehneilsen: An archive is not any old URI, but the address of the resource with the archive. It needs to have the expected directory layout (subdirecties with a given format and metadata yaml files with a given format in them), but it doesn't need to be an S3 bucket. I've been testing it on local disk with local files with URI's that look like `file:///my/test/archive/dir`. In principle, it could located at any URI that `lsst.resources` supports.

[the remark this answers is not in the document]
ehneilsen: I suspect it might not work with an http or https URI, though, because lsst.resources does not implement the whole API for http, and the code uses a part of the API that throws a NotImplementedError. So, only some kinds of URIs are supported.

schedview/collect/opsim.py
ehneilsen: Although `lsst.resources` lets you open a resource and provides an io handle that looks a lot like an opened file, this io object does not support random access for S3 buckets, and sqlite3 (unsurprisingly) uses random access. So, to read an opsim database file, I need to copy the remote resource to a local one. `lsst.resources` provides an `as_local` to create a local resources from remote ones in order to support random access where needed. (I'm pretty sure it just copies the file behind the scenes for you.) But, the `as_local` method on an S3 bucket does something with threads that doesn't play nice when run in a panel or bokeh server, so I added a branch to just make the local copy myself.
If `lsst.resources` added support for random access for S3 resources, I would expect this to be unnecessary. But, I think this comes down to what was practical with the S3 protocol, so this is not something I expect to happen.

schedview/data/sim_metadata.yaml
ehneilsen: Each simulation will have its own sim_metadata file; multiple simulations in a file isn't something that can happen.
This data file is just an absolute minimum needed for running the tests as they are. It will get expanded as new features get added (e.g. searching by date).
What this tells me is that I need to write a documentation page on the archive structure and how I plan for it to be used.

schedview/app/prenight/prenight.py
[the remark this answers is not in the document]
rhiannonlynne: Ok, thanks for explaining. I suspect some additional docstrings or comments in the code would help still.

schedview/collect/opsim.py
rhiannonlynne: Ah, right because "read" (such as from lsst.resources) doesn't read sqlite files ... 
Is this really a threading problem with bokeh/panel, or is it just 
 - we have a URI pointing to an opsim database output that is in an S3 bucket 
 - we want to read the database using sqlite, as it's a sqlite file
 - because we're using lsst.resources to access our S3 resources, we can't use sqlite
 
 (and can I just page @timj here for confirmation we're not missing something?)

schedview/data/sim_metadata.yaml
rhiannonlynne: Yes, it was just a bit hard figuring it out from the code. There is some information on what is written into the archive, but I also think there were some things we'd talked about previously that have maybe gotten outdated, and also there is a lot of additional information which is being written into the archive. 
Indeed, I find I'm not clear if "archive" is a single directory with a single run or something else.

ehneilsen: Yes, I've started writing a document on the archive structure. I don't think it needs to be a long document, but still too long to put in a comment, and it applies more generally than just a specific point in the code: someone who wants it shouldn't have to search the code to find it, and it would be bad to have to maintain the documentation in multiple places.
Not sure where the best place to put the document is; in the schedview docs, in the rubin_scheduler docs, and as a stand-alone tech note are all reasonable options, I think. From the point of view of writing it, I don't think it much matters: it's just an rst doc either way. But, it needs to go somewhere.
"Archive" is the directory tree populated with all the simulations and associated files: for each run, there is a directory in the archive with a metadata file and a collection of other files, at a minimum including the sqlite3 database of visits.

rhiannonlynne: Any of rubin_scheduler docs, schedview docs, or a tech note seem reasonable to me too. I suppose if the archive is created from rubin-scheduler, perhaps it makes the most sense to put the documentation of it there though (and hopefully having it in the docs and not a tech note means that it will be updated appropriately instead of forgotten).

rhiannonlynne: Perhaps an even better reason to put it in the rubin_scheduler docs then, since then it could be completely removed with future updates if needed.

schedview/collect/opsim.py
timj: The purpose of `as_local` is to download the remote file to the local file system so you can use it directly. I'm a bit confused as to how that messes you up given that you have to download the file anyhow. It's using boto3 to download the file which is what I assume you would be doing yourself.

`resource.open()` does return a file like object that supports seek and read. What do you need that it doesn't support?

ehneilsen: The docstring for `ResourcePath.open` states:

```
            `io.IOBase.seekable` is not guaranteed to be `True` on the returned
            object.
```

so I've been making sure only to use local files for random access.

ehneilsen: If `as_local` is not run in python main thread, it throws a `RuntimeError: cannot schedule new futures after interpreter shutdown`. Downloading the data with `resource_path.open(...).read()` seems to work fine in any thread.

When threading is on in a panel application, the code to support the dashboards is run in a thread, so `as_local` fails. So, an alternative to doing the download myself would be to force panel to not use threading, but I don't think I want to do this.

timj: It is true for S3 though. Have you tried it? cc @natelust who worked on this.

timj: Re `as_local`. Wow. All it's doing is calling `boto3.client.download_file` which I think uses multiple streams. This must be a general problem for everyone and not specific to `ResourcePath`.

https://github.com/lsst/resources/blob/main/python/lsst/resources/s3.py#L210

Whereas `read` uses `get_object`.

ehneilsen: I don't recall; I think I did try it initially with http (before I had a bucket I could use set up), and that failed. Once I realized that the sqlite3 connect API can only take actual paths to local files, not io objects, I realized that I needed a local file in any case.

timj: A quick scan on the stack overflow says that this error is caused if somewhere something is forgetting to call `mythread.join()`.

ehneilsen: Indeed I do not seem to be the first to have encountered something like this; I found this while debugging: https://stackoverflow.com/questions/67419800/error-cannot-schedule-new-futures-after-interpreter-shutdown-while-working-thr

ehneilsen: In this case, though, we don't *want* to join, because we want to keep different dashboard sessions running in different threads.

timj: Also this might be relevant: https://github.com/boto/boto3/issues/3113

It suggests adding:

```
from concurrent.futures import ThreadPoolExecutor
```
to the top of your code.

ehneilsen: At least for one instance where it's relevant, importing ThreadPoolExecutor seems to work. Thanks!

ehneilsen: Oops, nope, I was wrong. I had threading turned off in panel when I tested it. It's still failing with threading turned on, even if I do the import.
It would be less hacky to see if I can't get the other approach mentioned in that thread (turning off threading in boto) working anyway.

rhiannonlynne: Well .. .this sort of works, but there's still something a bit wrong because I lost all of the column names .. 

Screenshot 2024-01-10 at 1 52 27 PM

ehneilsen: A minor change to `S3ResourcePath` that lets threading be turned off by setting an environment variable seems to fix the issue, making `as_local` behave even when run in threads. @timj , I should make and issue and pull request on `lsst.resources`?

timj: Ok. Yes. Thanks.

rhiannonlynne (approved):

rhiannonlynne (approved):

PR #64, opened 2024-01-25
tickets/PREOPS-4829: make schedview robust w.r.t. surveys with present but empty observation lists

ehneilsen:
In previous versions of rubin_scheduler, instances of survey objects either didn't have an `observations` member at all, or had one with observations in it. In the latest version on main, it is sometimes present, but an empty list, which causes schedview to throw an exception. Make schedview handle that the same way as it would if there were no such member at all.


schedview/compute/scheduler.py
```diff
@@ -317,12 +317,16 @@ def make_unique_survey_name(scheduler, survey_index=None):
     except AttributeError:
         survey_name = str(survey)
 
-    if hasattr(survey, "observations") and (survey.survey_name != survey.observations["note"][0]):
+    if (
+        hasattr(survey, "observations")
+        and ("note" in survey.observations)
+        and (survey.survey_name != survey.observations["note"][0])
+    ):
         survey_name = f"{survey.observations['note'][0]}"
```
rhiannonlynne, line 325: Oh, this explains a thing maybe. 
The "note" isn't always what the survey name ought to be, and I had thought we'd done a better job at putting properly descriptive "survey names" into the Surveys. It's possible this isn't getting where it needs to be, but this might be worth checking again. 
In general, I don't think we want the note to be the survey name, and it should have been updated at the rubin_scheduler end by now. (I think we may also have "target" in observations, as well as "note" and it's possible that both target and note may be useful for use by schedview.).

rhiannonlynne: I think that's a separate thing than you wrote this ticket for, so will just recommend checking in on it again.

rhiannonlynne (approved):

schedview/compute/scheduler.py
ehneilsen: If I recall correctly, I originally used survey_name and only fell back on note when it was necessary, and the auxtel team asked me not to do this, but to always use note if it was present, even if survey_name was present and set to something else. So, I think changes to this will require coordinated changes to the auxtel scheduler configuration, not just the stuff in rubin_scheduler.

PR #65, opened 2024-01-25
tickets/PREOPS-4827: update test pickles to be compatible with current rubin_scheduler

ehneilsen:
recent changes on main have made the current set of scheduler pickles used for schedview testing incompatible.
This should not get merged until there's a new tag on rubin_scheduler.


rhiannonlynne:
Not sure if you need to redo this again, with the newly tagged rubin_scheduler.

rhiannonlynne (approved):
Not sure if you need to update again, but seems straightforward if so.

PR #67, opened 2024-01-31
PREOPS-4838 Limiting the datetime selector’s range to those for which there are sky brightness files

emanehab99:
Get the date range covered by SkyBrightness_Pre data files and limit date picker to this range so that user cannot select outside the values that have data


emanehab99:
Thanks for the feedback @ehneilsen. I've wrapped the code in a function and rebased the branch

emanehab99:
@ehneilsen also fixed the tests

ehneilsen (approved):

PR #70, opened 2024-02-16
PREOPS-4906 - Use scrolling rather than pagination to show overflowing content in rewards and basis function tables

emanehab99:
Remove pagination in both tables
set table height to enable scrolling
remove basis function class column
introduce column widths in basis function table
fix logo url


ehneilsen (approved):
The changes themselves look fine, but please use branch names following the conventions described here: https://developer.lsst.io/work/flow.html#dm-git-branching-policy
For this case, just rename the branch to tickets/PREOPS-4906 before merging.

PR #71, opened 2024-02-22
PREOPS-4906 - Use scrolling rather than pagination to show overflowing content in rewards and basis function tables

emanehab99:
Remove pagination in both tables
set table height to enable scrolling
remove basis function class column
introduce column widths in basis function table
fix logo url


schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -730,9 +730,10 @@ def create_summary_widget(self):
             disabled=True,
             selectable=1,
             hidden_columns=["tier", "survey_url"],
-            pagination="remote",
-            page_size=4,
+            # pagination="remote",
```
ehneilsen, line 733: Given that we have the old version in git, maybe just delete these lines instead of commenting them.

ehneilsen (approved):
Looks fine to me.

PR #73, opened 2024-02-23
PREOPS-4935: Improve logging

alserene:
Improve logging within the dashboard’s custom debugging panel and also in stderr.
- Add more logging messages throughout code.
- Send logging information to stderr (as well as dashboard debugging panel) using Python’s inbuilt logger module.
- Prevent debugging messages from spilling outside debugger area.
- Improve formatting of dashboard debugger.

### DIFF
--- a/schedview/app/scheduler_dashboard/scheduler_dashboard.py
+++ b/schedview/app/scheduler_dashboard/scheduler_dashboard.py
@@ -84,9 +84,28 @@ logging.basicConfig(
 )
 
 # Change styles using CSS variables.
-stylesheet = """
+h1_stylesheet = """
 :host {
---mono-font: Helvetica;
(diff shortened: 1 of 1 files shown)


ehneilsen (commented):
Is writing directly to stderr really the best way to handle this? Wouldn't using the python logger configured with an appropriate handler be more flexible if we need to do something else later?

alserene:
> Is writing directly to stderr really the best way to handle this? Wouldn't using the python logger configured with an appropriate handler be more flexible if we need to do something else later?

Sure, I can implement something like this.

ehneilsen (approved):
Looks good! This is exactly what I had in mind.

emanehab99 (approved):
Looks good! Thanks Alice

PR #74, opened 2024-02-23
PREOPS-4937: Reformat debugging panel.

alserene:
Overflow issue was found to be related to the debugging card being collapsed when first debugging messages were added. The problem is fixed by the card being uncollapsed from the start.

Some minor additional formatting changes were applied to make the uncollapsed debugger look nice, as well as stylesheets being defined for all headings in the dashboard.


emanehab99:
Nice work!

emanehab99 (approved):

PR #75, opened 2024-02-27
tickets/PREOPS-4948: functions to support imitating a DES nightsum report using a Times Square notebook

ehneilsen:
Several things will get refactored once some things get added as stackers in maf, but I want this done as a separate PR so the initial version doesn't rely on getting updated at the same time as rubin_sim.


schedview/collect/opsim.py
```diff
@@ -42,18 +150,53 @@ def read_opsim(opsim_uri, start_time="2000-01-01", end_time="2100-01-01"):
...
+    stackers : `list` [`rubin_sim.maf.stackers`], optional
+        Stackers to be used to generate additional columns.
+
+    Returns
+    -------
+    visits : `pandas.DataFrame`
+        The visits and their parameters.
+    """
+    ddf_field_names = tuple(ddf_locations().keys())
+    constraint = f"target IN {tuple(field_name for field_name in ddf_field_names)}"
```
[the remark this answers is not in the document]
rhiannonlynne: But Target is also good to use. Bear in mind it may change, too. (Peter is considering adding a sequence number to the current content).

schedview/compute/maf.py
```diff
@@ -0,0 +1,112 @@
...
+        this_constraint = f"filter == '{this_filter}'"
+        if len(constraint) > 0:
+            this_constraint += f" AND {constraint}"
+        slicer = maf.HealpixSlicer(nside=nside)
+        bundles[this_filter] = maf.MetricBundle(metric, slicer, this_constraint)
+
+    compute_metric(visits, bundles)
+    metric_values = {b: bundles[b].metric_values for b in bundles if bundles[b].metric_values is not None}
+
+    return metric_values
```
rhiannonlynne, line 112: I think I agree that most of this will probably go elsewhere (such as into MAF, then call the batch, etc. or for an advanced use, to just call MAF directly).

schedview/compute/visits.py
```diff
@@ -0,0 +1,227 @@
...
+        day_obs_mjd, and day_obs_iso8601.
+    """
+    day_obs_mjd = np.floor(visits["observationStartMJD"] - 0.5).astype("int")
+    day_obs_datetime = Time(day_obs_mjd, format="mjd").datetime
+    day_obs_date = [datetime.date(t.year, t.month, t.day) for t in day_obs_datetime]
+    day_obs_iso8601 = tuple(str(d) for d in day_obs_date)
+    visits.insert(1, "day_obs_mjd", day_obs_mjd)
+    visits.insert(2, "day_obs_date", day_obs_date)
+    visits.insert(3, "day_obs_iso8601", day_obs_iso8601)
+    return visits
```
rhiannonlynne, line 32: We might want to go ahead and also incorporate the summit tools as long they don't require the stack to install. And use Merlin's computation of DAYOBS there. Then we would always stay in sync with the same definition of DAYOBS, in case it changes (due to taking flats or other calibration data in the morning, for example).

schedview/compute/visits.py
```diff
@@ -0,0 +1,227 @@
...
+        The modified DataFrame with a t_eff column.
+    """
+    if nighbor_column is None:
+        col_index = len(visits.columns)
+    else:
+        col_index = tuple(visits.columns).index(nighbor_column)
+
+    value = schedview.compute.compute_metric_by_visit(visits, metric, constraint)
+    visits.insert(col_index, column_name, value)
+    return visits
```
rhiannonlynne, line 79: Right .. we were going to turn Teff into a stacker ..

schedview/compute/visits.py
```diff
@@ -0,0 +1,227 @@
...
+        + visits["visitTime"]
+        - visits["visitExposureTime"]
+    )
+    slew_time_col_index = tuple(visits.columns).index("slewTime")
+    visits.insert(slew_time_col_index + 1, "overhead", overhead)
+
+    filter_col_index = tuple(visits.columns).index("filter")
+    previous_filter = visits["filter"].shift(1)
+    visits.insert(filter_col_index + 1, "previous_filter", previous_filter)
+    return visits
```
rhiannonlynne, line 108: And now I'm wondering what happens regarding focus frames and if the AOS system needs additional time.

rhiannonlynne (approved):
Like we talked about, there are things to rearrange later, but this is a good start.

PR #77, opened 2024-03-08
tickets/PREOPS-4975: provide matplotlib implementation of hpix and visit maps

ehneilsen:


rhiannonlynne:
Can you paste in some examples of the plots that are created? 
I wasn't really expecting a new dependency and an entirely new way of creating plots, for things which I'm assuming are equivalent to the outputs of metric bundles, so it would be helpful to see what the results look like and what added functionality they're providing over the plotting utilities that are already in MAF.  (especially since this would be adding a new 'look and feel' kind of plot output, I'd just like to get a handle on what this looks like).

rhiannonlynne (commented):
(see comment in conversation)

ehneilsen:
> Can you paste in some examples of the plots that are created? I wasn't really expecting a new dependency and an entirely new way of creating plots, for things which I'm assuming are equivalent to the outputs of metric bundles, so it would be helpful to see what the results look like and what added functionality they're providing over the plotting utilities that are already in MAF. (especially since this would be adding a new 'look and feel' kind of plot output, I'd just like to get a handle on what this looks like).

It looks almost the same as the bokeh plot, just without the interactive elements:
plotgridexample

ehneilsen:
It isn't there in this example, but I did also write functionality that adds the limits of what can be observed on a night (based on max zenith distance, sunrise, and sunset times).

ehneilsen:
An alternative might have been to use Eli and Alex's [skyproj](https://skyproj.readthedocs.io/en/latest/).

ehneilsen:
I just reimplemented this using skyproj instead of cartopy, and indeed it is much simpler. I'm finishing up the unit tests now. When I do the refactor to get stuff related to nightsum into maf, it may be good to make this a maf plotter (which should be really easy).

ehneilsen:
BTW, I did look at using the existing maf plotters, but as far as I can see, it isn't possible to overplot things like the ecliptic or polygons on top of healpix maps. It also looks like the code in that does things like plotting the ecliptic (BaseSkyMap) relies on Basemap. I used to use Basemap, but back when I was working on DES, the last of Basemap maintainers said he wouldn't be maintaining it anymore, and was pushing to people to use cartopy. This is why I've been using cartopy since then. (Also, once I learned to use it, I prefer cartopy to Basemap, but really have had headaches with both.) So, I worry about the long-term viability of the code that uses Basemap, and am indeed a little surprised it still works at all, given how long ago Basemap was declared depricated by its maintainers.

I've looked for the announcement, but it seems to be gone now. The closest I could find was [here](https://github.com/SciTools/cartopy/issues/920#issuecomment-321329691), where the maintainer of Basemap states that he declared it deprecated and provides a link to his announcement, but the link is broken. But, Basemap still seems to be working. Maybe a new team picked up maintaining it?

ehneilsen:
BTW, this is what the skyproj version of the maps looks like, with the zd=70° circles and morning and evening twilight restored. These are useful for showing the limits of what can be seen on the night in question.
skyprojmaps

rhiannonlynne (approved):
Looks good. 

At some point, we're going to have to grapple with the fact that your code is all pandas-based and almost all of the scheduler and rubin-sim are numpy arrays. I know they're convertible, but converting back and forth seems inefficient.

PR #82, opened 2024-03-21
tickets/PREOPS-4949: Move computations of some visit params to rubin_sim stackers

ehneilsen:
This PR also include changes for PREOPS-4990, which let the scheduler snapshot dashboard run without rubin_sim installed.


rhiannonlynne (approved):
Looks good, sorry for the delay.

PR #84, opened 2024-04-17
tickects/PREOPS-5067: Add basis function and survey rewards timeline plots that do not use python callbacks

ehneilsen:
Needed to support adding such plots to the Times Square pre-night briefing.


rhiannonlynne:
My look at this was entirely cursory, as most of the move from javascript to python callbacks is beyond what I'm familiar with. 
As ever, I would prefer more comment lines to help understand why a particular piece of code is being used (why one version of the start date instead of another, for example).

rhiannonlynne (approved):

PR #85, opened 2024-04-23
tickets/PREOPS-4750: Support selecting data from the EFD in schedview's scheduler snapshot dashboard

emanehab99:
This PR adds a third operating mode for the on-the-fly scheduler dashboard that works only in USDF environment
To enable this mode, run the dashboard using --usdf command line parameter

The dashboard utilises LSST EFD Client to query USDF S3 bucket using a certain datetime and telescope (main, auxtel or both). Snapshots with a timestamp between the start of the night and selected datetime are listed most recent to least.
Data is loaded from snapshot file using the LSST resources module.


schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -566,7 +573,7 @@ def _update_color_palette(self):
     # ------------------------------------------------------- Internal workings
 
     def clear_dashboard(self):
-        """Clear the dashboard for a new pickle or a new date."""
+        """Clear the dashboard for a new pickle or a new date"""
```
ehneilsen, line 576: We should keep the period.

schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -1449,6 +1463,54 @@ def __init__(self, data_dir=None):
             self.param["scheduler_fname"].update(path=f"{data_dir}/*scheduler*.p*")
 
 
+class USDFScheduler(Scheduler):
```
ehneilsen, line 1466: We should call it something like `LFAScheduler` rather than `USDFScheduler` (and we should consider renaming all of these to something less ambiguous, like `LFASchedulerSnapshotDashboard`, so it doesn't get confused with the actual scheduler, or other tools relate to the scheduler). The LFA used might not always be at the USDF (e.g. it could be at the summit), and we may use other access methods at the USDF.

schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -1508,14 +1582,51 @@ def scheduler_app(date_time=None, scheduler_pickle=None, **kwargs):
...
+        # set specific widget props for data loading parameters
+        # in URL and restricted modes
         data_loading_widgets = {
             "scheduler_fname": {
-                # "widget_type": pn.widgets.TextInput,
                 "placeholder": "filepath or URL of pickle",
             },
             "widget_datetime": pn.widgets.DatetimePicker,
         }
+    # Load pickles from USDF S3 bucket
```
ehneilsen, line 1593: This should work fine for S3 buckets other than the USDF ones.

schedview/app/scheduler_dashboard/scheduler_dashboard.py
```diff
@@ -1669,11 +1779,20 @@ def parse_arguments():
         help="Let the user specify URLs from which to load data. THIS IS NOT SECURE.",
     )
 
+    parser.add_argument(
+        "--usdf",
```
ehneilsen, line 1783: Switch from `--usdf` to `--lfa`.

ehneilsen (approved):
Only minor changes requested -- overall it looks good, and works great when I test it.

emanehab99:
Comments resolved. Thanks Eric

PR #88, opened 2024-05-07
tweaks to text in the scheduler snapshot dashboard UI

ehneilsen:
Responses to feedback since making the post-user testing version available at the USDF.


emanehab99 (approved):
Looks good

PR #93, opened 2024-06-17
PREOPS-5239: Switch schedview from using bokeh.plotting.figure.circle to bokeh.plotting.figure.scatter with marker=circle

ehneilsen:
bokeh.plotting.figure.circle is sometimes throwing deprecation warnings, and sometimes silently just not plotting points (with the latest version of bokeh).


rhiannonlynne (approved):
It will be interesting to see if bokeh has significant changes in the API, and how we can keep up with it, but - this looks fine.

PR #96, opened 2024-07-12
tickets/PREOPS-5288: Add hovertool giving visit data to visit map figures used in the pre-night briefing.

ehneilsen:


rhiannonlynne:
This looks like it also consolidated a bunch of other code so that the planisphere and armillary sphere are now just one piece of code? Nice.

rhiannonlynne (approved):

PR #100, opened 2024-08-14
tickets/PREOPS-5367: use target_note instead of target, and make schedview more robust to changes in the opsim database schema

ehneilsen:


rhiannonlynne:
This looks fine. Reran CI manually to ensure picking up rubin_scheduler v1.3.0 - https://github.com/lsst/schedview/actions/runs/10822668915

rhiannonlynne:
Ah - somewhere you are instantiating a basis function that has changed its API. Unit test still fails.

rhiannonlynne (changes requested):
unit test failure needs attention

ehneilsen:
Rebasing onto main after the merge of PREOPS-4102 seems to have solved the test failures. I don't know how this should up as something in SunHighLimitBasisFunction though.

rhiannonlynne:
This was probably that: 
the pickle used SunHighLimitBasisFunction. In rubin_scheduler, we redirected SunHighLimitBasisFunction to CloseToTwilightBasisFunction (partly because of the confusion the testers had about the name, to make it clear that the goal was to get close to twilight).  So the pickle was finding a valid class, but all of the attributes it restored were wrong (because they were actually for SunHighLimitBasisFunction) so it didn't actually work. 
Now you have new pickles, it should be fine.

rhiannonlynne (approved):

PR #102, opened 2024-08-23
tickets/PREOPS-4102: Support customizing which columns are shown in schedview's Times Square reports

ehneilsen:
There will be a corresponding pull request to schedview_notebooks shortly.


rhiannonlynne:
I don't really know enough about bokeh to know if the changes here are implemented as they should -- I looked at the notebook and it seems that it does what I'd expect (although I kind of thought I should be able to set the order in which values showed up, if I got to pick what they were ..). 
Ruff problem, but approved.

rhiannonlynne (approved):

PR #108, opened 2024-09-30
tickets/PREOPS-5539: Highlight winning survey on snapshot dashboard

alserene:
To highlight which survey is "chosen" for a particular snapshot, a dynamic label has been created with the text "Chosen survey is ... ". This label is updated to display the "chosen" tier and survey whenever a new snapshot is loaded.

The label sits above the Scheduler Summary table header on a dark teal background.

chosen survey label


ehneilsen (approved):
The changes in this commit look good, but a prior commit broke the LFA version. I don't see why that should stop this pull request, though.

PR #110, opened 2024-10-02
tickets/SP-1595: restore panel dependency of snapshot list on telescope and date

ehneilsen:


emanehab99 (approved):
Looks good. Thanks for figuring this out

PR #112, opened 2024-10-07
tickets/SP-1624: Update schedview to work with rubin_scheduler v3.0.0rc0

ehneilsen:


schedview/compute/scheduler.py
```diff
@@ -87,11 +25,13 @@ def replay_visits(scheduler, visits):
     visits : `pandas.DataFrame`
         A table of visits to add.
     """
-    for visit_id, visit_record in visits.iterrows():
-        obs = _make_observation_from_record(visit_record)
-        scheduler.add_observation(obs)
+    obs = SchemaConverter().opsimdf2obs(visits)
+    for this_obs in obs:
+        scheduler.add_observation(this_obs)
```
rhiannonlynne, line 30: It would be interesting to check that if you just add them all in an array (i.e. 
`scheduler.add_observation_array(obs)`)
that this all works as expected. It should, and it should be faster.

rhiannonlynne (approved):

rhiannonlynne:
The failing unit test just looks like it needs a rerun of the generation of the pickle data. 
There was a name-change in the keyword argument for the AltAzShadowMaskBasisFunction, such that `altaz_limit_pad` just became `pad`.

PR #113, opened 2024-10-16
tickets/SP-1629: update tests to work with post-SP1629 rubin_scheduler

ehneilsen:


yoachim (approved):
Looks good. I did swap `altaz_limit_pad` to be just `pad`, so once that's updated the unit test should pass.

PR #114, opened 2024-10-28
tickets/SP-1667: Merge the Times Square OR4 nightsum with the nightly nightsum

ehneilsen:
Add wrapper code that can figure out how to load what the user asks for, whether it is a sim or data from an instrument in consdb.

### DIFF
--- a/schedview/collect/consdb.py
+++ b/schedview/collect/consdb.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import rubin_sim.maf
+from rubin_scheduler.scheduler.utils import ObservationArray
 from rubin_scheduler.utils.consdb import load_consdb_visits
 
 
@@ -11,7 +12,12 @@ def read_consdb(
     **kwargs,
 ) -> pd.DataFrame:
     consdb_visits = load_consdb_visits(instrument, *args, **kwargs)
-    visit_records: np.recarray = consdb_visits.opsim.to_records()
+    # If the return is empty, to_records won't be able to determine the
+    # right dtypes, so use ObservationArray()[0:0] instead to make the
+    # empty recarray with the correct dtypes.
+    visit_records: np.recarray = (
+        consdb_visits.opsim.to_records() if len(consdb_visits.opsim) > 0 else ObservationArray()[0:0]
+    )
     for stacker in stackers:
         visit_records = stacker.run(visit_records)
     visits = pd.DataFrame(visit_records)
--- a/schedview/collect/visits.py
+++ b/schedview/collect/visits.py
@@ -1,0 +1,90 @@
+from warnings import warn
+
+import pandas as pd
+from astropy.time import Time
+from lsst.resources import ResourcePath
+from rubin_scheduler.utils import ddf_locations
+from rubin_sim import maf
+
+from schedview import DayObs
+
+from .consdb import read_consdb
+from .opsim import read_ddf_visits as read_ddf_visits_from_opsim
+from .opsim import read_opsim
+
+KNOWN_INSTRUMENTS = ["lsstcomcamsim"]
+
+# Use old-style format, because f-strings are not reusable
+OPSIMDB_TEMPLATE = (
+    "/sdf/group/rubin/web_data/sim-data/sims_featureScheduler_runs{sim_version}/baseline/"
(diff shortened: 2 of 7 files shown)


schedview/collect/visits.py
```diff
@@ -65,6 +65,9 @@ def read_ddf_visits(*args, **kwargs) -> pd.DataFrame:
     all_visits = read_visits(*args, **kwargs)
 
     ddf_field_names = tuple(ddf_locations().keys())
+    # Different versions of the schedule include a DD: prefix, or not.
+    # Catch them all.
+    ddf_field_names = ddf_field_names + tuple([f"DD:{n}" for n in ddf_field_names])
```
rhiannonlynne, line 69: In new baselines, the Roman field shows up as "DD: RGES" .. although it's not really a DDF.
This is probably something to fix in the rubin_scheduler outputs, but it's worth noting here too. 
Creating similar kinds of plots for the Roman field is likely not terrible.

schedview/dayobs.py
```diff
@@ -171,8 +165,7 @@ def from_time(
         if dayobs_datetime.tzinfo is None:
             dayobs_datetime = dayobs_datetime.replace(tzinfo=datetime.timezone.utc)
 
-        dayobs_datetime = dayobs_datetime.astimezone(DAYOBS_TZ)
-        dayobs_date = datetime.date(dayobs_datetime.year, dayobs_datetime.month, dayobs_datetime.day)
+        dayobs_date = dayobs_datetime.astimezone(DAYOBS_TZ).date()
```
rhiannonlynne, line 168: What is DAYOBS_TZ ? 
I'm just curious because dayobs is essentially defined in UTC and I suspect we don't want to mess with that too much ... otherwise people in vastly different timezones like Australia might end up locating visits on completely different days (or at least having confusing "day" labels).

schedview/collect/consdb.py
```diff
@@ -11,7 +12,12 @@ def read_consdb(
     **kwargs,
 ) -> pd.DataFrame:
     consdb_visits = load_consdb_visits(instrument, *args, **kwargs)
-    visit_records: np.recarray = consdb_visits.opsim.to_records()
+    # If the return is empty, to_records won't be able to determine the
+    # right dtypes, so use ObservationArray()[0:0] instead to make the
+    # empty recarray with the correct dtypes.
+    visit_records: np.recarray = (
+        consdb_visits.opsim.to_records() if len(consdb_visits.opsim) > 0 else ObservationArray()[0:0]
+    )
```
rhiannonlynne, line 20: If it's empty, does anything else need to run?

rhiannonlynne (approved):
Some questions, but mostly for clarification. 
I am curious though if dayobs means the same thing (in UTC) even if the timezone used is something like Australia.

schedview/dayobs.py
ehneilsen: This is me being punctilious about following SITCOMTN-032 exactly, which states that dayObs is:

> the observation day (in timezone UTC-12:00) when the takeImages command began executing. 

which line 10 of `dayobs.py`:

`DAYOBS_TZ = datetime.timezone(datetime.timedelta(hours=-12))`

schedview/collect/consdb.py
ehneilsen: If I don't know exactly what the result will be used for, I find it often helps save trouble later on if I return an empty tables (whether represend as arecarray or DataFrame) with the expected columns and column types, so that it can be indexed or sliced. For example, if someone wants to count the number of visits with cos(fieldRA)>0.5 (I don't know why they might want to do this, it's just an example), they might do:

`print(len(visits.loc[np.cos(visits.fieldRA>0.5), :])`

If it is an empty DataFrame with a fieldRA of a type you can take a cosine of, this will work fine whether or not visits has any rows. But, if its None or a DataFrame with no fields defined, or if, say fieldRA is defined but a string, it will fail. In this case, though, there's no reason why it should fail. The same sort of thing happens with recarrays.

schedview/dayobs.py
rhiannonlynne: I see, so it's not that DAYOBS_TZ can change, it's also defined.

schedview/collect/consdb.py
rhiannonlynne: Sounds good.

PR #117, opened 2024-11-14
ticktes/SP-1735: use scheduler name when possible

ehneilsen:
Look at the comment on line 262 for an explanation of where this came from to begin with.

### DIFF
--- a/schedview/compute/scheduler.py
+++ b/schedview/compute/scheduler.py
@@ -265,11 +265,13 @@ def make_unique_survey_name(scheduler, survey_index=None):
     # attribute.
     try:
         observation_note = f"{survey.observations['note'][0]}"
+        if len(observation_note) == 0:
+            observation_note = None
     except (AttributeError, ValueError, TypeError):
         observation_note = None
 
     if (observation_note is not None) and (survey.survey_name != observation_note):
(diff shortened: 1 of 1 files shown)


schedview/compute/scheduler.py
```diff
@@ -265,11 +265,13 @@ def make_unique_survey_name(scheduler, survey_index=None):
...
     try:
         observation_note = f"{survey.observations['note'][0]}"
+        if len(observation_note) == 0:
+            observation_note = None
     except (AttributeError, ValueError, TypeError):
         observation_note = None
 
     if (observation_note is not None) and (survey.survey_name != observation_note):
-        survey_name = f"{survey.observations['note'][0]}"
+        survey_name = f"{survey.survey_name} ({survey.observations['note'][0]})"
```
rhiannonlynne, line 274: You might find that scheduler note is being used instead of note (if observation is the ObservationArray).

rhiannonlynne (approved):

schedview/compute/scheduler.py
ehneilsen: I've modified the code to use `scheduler_note` if it is present, but to fall back on `note` if it is not.

PR #119, opened 2024-12-02
tickets/SP-1755: add ephemeris data to schedview.DayObs

ehneilsen:
This implementation does not depend on `rubin_scheduler,` or other parts of `schedview`, just `astropy.coordinates`, `astropy.time`, and `scipy.optimize`. A version that used the `rubin_scheduler` almanac would be faster: the computation of positions with astropy is slow, and the minimizer calls it many times. If you're just concerned with one night, it's not slow enough to be bothersome, but if you're doing many nights, it will be a bottleneck.

Maybe if this is ever extracted I'll but a derivative of it it `schedview` that overrides these methods with `rubin_scheduler` almanac lookups.

Comparing the alts at the times of both DayObs and the almanac to requested alts (using astropy.coordinate.get_body in vaccuum), the DayObs implementation has a mean difference of -0.03" in alt with an RMS of 2.6", while the almanac times give a mean difference of -1.1" with an RMS of 5.5". So, they agree both with each other and astropy (in vaccuum) to within a few arcseconds.


ehneilsen:
I just checked in a tweak to the optimization that results in an RMS difference in the DayObs alt with a mean difference of 0.004 arcseconds and an RMS of 0.16 arcseconds.

schedview/dayobs.py
```diff
@@ -1,16 +1,26 @@
...
 
 DAYOBS_TZ = datetime.timezone(datetime.timedelta(hours=-12))
 MJD_EPOCH = datetime.date(1858, 11, 17)
 ONE_DAY = datetime.timedelta(days=1)
 
+LSE30_ATMOSPHERE = {
+    "pressure": 750.0 * 100 * u.Pa,
+    "temperature": 11.5 * u.deg_C,
+    "relative_humidity": 0.4,
+}
```
rhiannonlynne, line 22: The `rubin_scheduler.utils.Site` class, when instantiated for "LSST" should also have this same information. If you need to keep this here separately, that's fine, but consider using the Site class instead so that if we find that a different value is more appropriate for Rubin, we can change it in as few places as possible.

schedview/dayobs.py
```diff
@@ -31,15 +41,32 @@ class DayObs:
...
     """
 
     # Use the standard library date as a lowest common denominator
     # that cannot represent more or less than we want it to.
     date: datetime.date
     int_format: IntDateFormat = "auto"
+    atmosphere: dict | None = None
+    location: EarthLocation = EarthLocation.from_geodetic(
+        lon=-70.7494 * u.deg, lat=-30.2444 * u.deg, height=2650.0 * u.meter
+    )
```
rhiannonlynne, line 60: I'd also be curious if this is different from the EarthLocation determined from Site("LSST").to_earth_location()

tests/test_dayobs.py
```diff
@@ -54,3 +66,86 @@ def test_dayobs(self):
...
+    def test_against_almanac(self):
+        num_nights_tested = 2
+        almanac_times = (
+            pd.DataFrame(ALMANAC.sunsets)
+            .sample(num_nights_tested, random_state=42)
+            .set_index("sunset", drop=False)
+        )
+        for sunset_mjd, night_times in almanac_times.iterrows():
+            day_obs = DayObs.from_time(sunset_mjd)
+            assert np.isclose(night_times.sunset, day_obs.sunset.mjd)
```
rhiannonlynne, line 129: I'm not sure the defaults for np.isclose are set up to catch the kinds of differences we'd like to catch, when being used on values that are on the order of 60980.0 .. 
I could very well be missing something, because this seems crazy but .. 
```
s = SURVEY_START_MJD
s2 = s + 3/24
rtol=1e-05
atol=1e-08
print(s, s2, (np.abs(s - s2)), (atol + rtol * np.abs(s2)), (atol + rtol * np.abs(s2))*24)
np.isclose([s], [s2])
```
indicates that this would match for differences up to 14 hours.

rhiannonlynne (commented):
I'm pretty sure that your calculated sunset/sunrise times are actually close to the times from the Almanac, but the unit test here is not set up to actually test this. 
Could you please update the unit test by either modifying the parameters for np.isclose or using something like almostEqual? 

I was also wondering about avoiding the use of Site('LSST') .. is this so that the class could be more generally used for other locations?

schedview/dayobs.py
ehneilsen: I originally used `rubin_scheduler.utils.Site` when I first wrote this, but I thought the goal was to move this somewhere where it wouldn't depend on rubin_scheduler, so I got rid of that dependency by copying it.

schedview/dayobs.py
ehneilsen: This is almost a cut and paste from Site from rubin_scheduler, again because that's what I originally used by was trying to remove the dependency.

tests/test_dayobs.py
ehneilsen: I just revised the test to set rtol=0 and atol = 1.0/(24*60), and did some experiments to verify it does what I expect.

schedview/dayobs.py
ehneilsen: I also made the modifications to the isclose calls to set rtol and atol to something reasonably demanding.

rhiannonlynne (approved):

PR #120, opened 2024-12-06
tickets/PREOPS-1708: Add narrative text message support

ehneilsen:


ehneilsen:
This PR uses the almanac based changes to DayObs to get the rise and set, which will get replaced when the other PR gets merged.

schedview/collect/nightreport.py
```diff
@@ -0,0 +1,120 @@
+import datetime
+from typing import Literal
+
+import requests
+
+from schedview.dayobs import DayObs
+
+MAX_RETRIES = 1
```
rhiannonlynne, line 8: MAX_RETRIES should be 2, based on my experience with the logging endpoints.

schedview/collect/nightreport.py
```diff
@@ -0,0 +1,120 @@
...
+        response = requests.get(api_endpoint, params)
+        try_number += 1
+
+    return response.json()
+
+
+def get_night_report(
+    day_obs: DayObs | str | int,
+    telescope: Literal["AuxTel", "Simonyi"],
+    api_endpoint: str = "https://usdf-rsp-dev.slac.stanford.edu/nightreport/reports",
```
rhiannonlynne, line 31: We need a better way to handle these endpoints. 
Presumably these are tools that should run at summit or at USDF - these services are available in both. 
The base URL differs, but the endpoint "nightreport/reports" is the same.
I don't have a super awesome suggestion at present, but one option might be something based on:
```
def get_clients() -> dict:
    """Return site-specific client connections. 

    Returns
    -------
    endpoints : `dict`
        Dictionary with `efd`, `obsenv`, 
        `narrative_log`, and `exposure_log`
        connection information.
        For the obsenv, narrative log and exposure log, these are only
        defined for the summit or USDF.
    """
    site = getSite()
    # This will fail if site is "unknown"
    # although could go with usdf_efd ..
    efd_client = makeEfdClient()
    if site == "summit":
        API_BASE = "https://summit-lsp.lsst.codes/"
        obsenv_client = EfdClient('summit_efd', db_name='lsst.obsenv')
    elif site == "tucson" or site == "base":
        # I don't know what the logging looks like here, 
        # but suspect it doesn't exist
        obsenv_client = None
        API_BASE = None
    else:
        obsenv_client = EfdClient('usdf_efd', db_name='lsst.obsenv')
        API_BASE = "https://usdf-rsp.slac.stanford.edu/"
    narrative_log_url =  API_BASE + "narrativelog/messages"
    exposure_log_url = API_BASE + "exposurelog/messages"
    return {'efd': efd_client, 'obsenv': obsenv_client, 
            'narrative_log': narrative_log_url, 'exposure_log': exposure_log_url}
```
although I think in practice, we don't want to use summit_utils' makeEfdClient tool for getting the EfdClient, for example (since it fails outside of any of the known places) and if we drop that, we may want to just drop getSite too.

schedview/plot/nightreport.py
```diff
@@ -0,0 +1,233 @@
...
+        print("", file=report)
+
+    if "confluence_url" in night_report and len(night_report["confluence_url"]) > 0:
+        night_plan_block = (
+            "BLOCK" + urllib.parse.urlparse(night_report["confluence_url"]).fragment.split("BLOCK")[-1]
+        )
+        if night_plan_block == "BLOCK":
+            night_plan_block = night_report["confluence_url"]
+        print(f"{headhash}# Night plan", file=report)
+        print(f"[{night_plan_block}]({night_report['confluence_url']})", file=report)
```
rhiannonlynne, line 49: You might want to make these actual clickable URLs - 
something like 
``` 
display(Markdown(f'Night plan : <a href="{url}" target="_blank" rel="noreferrer noopener">{night_plan_block}</a>'))
```
(translated to print into your stringIO) could  work. 
.. I've since found that if these URLs point to zephyr scale, they work, but otherwise something in TimesSquare or the USDF RSP is trapping them internally (I think there may be supposed to be a proxy to JIRA that hijacks the URL .. JIRA URLs  or github URLs don't work, but zephyr scale ones do) - but it's worth a try, and maybe we can figure out what's stopping them from opening properly.

rhiannonlynne (commented):
Some comments here - I think you might still be finishing this up a little, which is fine - just tag me in the "conversation" when you want me to have another look. 
If you could show an example of the outputs, that'd be great (I think you have this already in a notebook -- just a screenshot would be fine). Mostly because I think much of the narrative log is currently unused (subsystems, etc) so if we're taking up a lot of real estate with the unused data, we might drop it. But I suspect that you're putting the results up in a way that makes them look "small" so it should be fine.

rhiannonlynne:
Oh - one more comment: 
this is how I'm getting narrative log messages in the evaluate script queue notebook and I'd like to know how these would integrate together : 
```

def get_narrative_log(t_start: Time, t_end: Time, narrative_log_endpoint: str) -> pd.DataFrame:
    """Get the narrative log entries."""    
    log_limit = 50000
    params = {"is_human" : "either",
              "is_valid" : "true",
              "has_date_begin" : True,
              "min_date_begin" : t_start.to_datetime(),
              "max_date_begin" : t_end.to_datetime(),
              "order_by" : "date_begin",
              "limit": log_limit, 
             }
    messages = query_logging_services(narrative_log_endpoint, params)
    # Modify narrative log content to match dataframes from errors and watcher topics better.
    # Strip out repeated \n\n and \r\n characters for nicer printing in dataframe.
    if len(messages) > 0:
        def strip_rns(x):
            return x.message_text.replace("\r\n", "\n").replace("\n\n", "\n").rstrip("\n")
        def make_time(x, column):
            return Time(x[column], format='isot', scale='tai').utc.datetime
        def clarify_log(x):
            if x.components is None:
                component = "Log"
            else:
                component = "Log " + " ".join(x.components)
            return component
        # Strip excessive \r\n and \n\n from messages
        messages['message_text'] = messages.apply(strip_rns, axis=1)
        # Add a time index
        messages['time'] = messages.apply(make_time, args=["date_begin"], axis=1)
        messages.set_index('time', inplace=True)
        messages.index = messages.index.tz_localize("UTC")
        # Join the components and add "Log" explicitly
        messages['component'] = messages.apply(clarify_log, axis=1)
        # rename some columns to match error data
        messages.rename({'time_lost_type': 'error_code', 'user_id': 'origin'}, axis=1, inplace=True)
        # Add a salindex so we can color-code based on this as a "source"
        messages['salIndex'] = 0
        messages['error_code'] = 0
        messages['finalStatus'] = "Log"
        messages['timestampProcessStart'] = messages.apply(make_time, args=["date_begin"], axis=1)
        messages['timestampRunStart'] = messages.apply(make_time, args=["date_added"], axis=1)
        messages['timestampProcessEnd'] = messages.apply(make_time, args=["date_end"], axis=1)
    print(f"Found {len(messages)} messages in the narrative log")
    if len(messages) == log_limit:
        print(f"Whoops, likely lost some log messages due to limit of {log_limit}.")
    return messages
  ```
  
I'm not sure if you'd considered running the narrative log for multiple nights or just one night, and thus potentially hitting the end of the list. 
And I'm not sure if a workable idea would be to use your utility to get the nightlog messages, and then reformat it as it is here, in a different function, so as to make it something that can fit into the remainder of the ScriptQueue, etc.

schedview/collect/nightreport.py
ehneilsen: Done

schedview/collect/nightreport.py
ehneilsen: We could figure out where it's running with something like `'.'.join(socket.getfqdn().split('.')[1:])` and have a dictionary somewhere that maps domains to base URLs, but maybe that should be a different PR?
What prompted me to make this PR now, even though it is a bit rough, is that it's getting to large, with too big a variety of stuff in it for one PR. I'd rather merge what I have so far and incrementally improve it than develop this onto one big monster PR.

schedview/plot/nightreport.py
ehneilsen: It is already a clickable URL: note the square braces and parenthesis, which is markdown for making a link.
But, it doesn't have your nice `target="_blank" rel="noreferrer noopener"` params, which is an improvement, so I just checked in the change.

ehneilsen:
> Oh - one more comment: this is how I'm getting narrative log messages in the evaluate script queue notebook and I'd like to know how these would integrate together :

...

> I'm not sure if you'd considered running the narrative log for multiple nights or just one night, and thus potentially hitting the end of the list. And I'm not sure if a workable idea would be to use your utility to get the nightlog messages, and then reformat it as it is here, in a different function, so as to make it something that can fit into the remainder of the ScriptQueue, etc.

There's *lots* more I want to do here. This set of narrative log messages are being used to support the timeline plot, which is currently just one timeline with all the same colored points with the same shape, and very directly formatted text is the hover text. I eventually want to color and shape code the points (both with the same meaning, to support the color blind), maybe split it into multiple parallel timelines, and add other events from the EFD as timelines on the same plot.
But, I think these are all incremental improvements that should go in different PRs.

Some of the formatting stuff will overlap what you need to do, but they are different use cases: we may want to tune it differently depending on whether it shows up as hover text or in a table. I think it may be more flexible not to try to reuse the same code for both, but we should certainly take inspiration from each other on this.

ehneilsen:
To see what this is all in aid of, see the "Night report ..." and the left pane of the "Value timelines" sections of this version of the night summary: https://usdf-rsp-dev.slac.stanford.edu/times-square/github/lsst/schedview_notebooks/nightly/scheduler-nightsum?day_obs=2024-12-05&visit_origin=lsstcomcam&ts_hide_code=1

rhiannonlynne:
Looks great. 
A separate ticket, but can the maps in the notebook you linked above show up in 
ugrizy-other
order?

rhiannonlynne (approved):

PR #121, opened 2024-12-09
tickets/SP-1742: Expand schedview documentation with examples

ehneilsen:
Not every report element is shown in the examples, but I think this is as many as are useful (or at least worth the effort) at the moment.


rhiannonlynne (approved):
Looks great.

PR #123, opened 2024-12-10
tickets/SP-1792: Pin rubin_scheduler to 3.4.0

ehneilsen:
Once this is merged, I can tag a new schedview (0.16.0) and create a new container, then push to phalanx.
This will sweep up a few more changes to schedview into 0.16.0, but I think that's okay.


rhiannonlynne (approved):
And then with this, if we update rubin_scheduler version, we tag another version of schedview, correct?

PR #124, opened 2025-01-06
tickets/SP-1789: plot EFD logevents on an event timeline plot

ehneilsen:
This PR is for the initial infrastructure and a few initial examples.
I expect plots to be supplemented and refined in additional issues and PRs.

### DIFF
--- a/schedview/collect/efd.py
+++ b/schedview/collect/efd.py
@@ -4,6 +4,7 @@ from collections.abc import Iterable
 from functools import partial
 
 import pandas as pd
+from lsst.summit.utils.efdUtils import makeEfdClient
 from lsst_efd_client import EfdClient
 
 from schedview.dayobs import DayObs
@@ -21,7 +22,7 @@ def _get_efd_client(efd: EfdClient | str | None) -> EfdClient:
         case str():
             efd_client = EfdClient(efd)
         case None:
-            efd_client = EfdClient(DEFAULT_EFD)
+            efd_client = makeEfdClient()
         case _:
             raise ValueError(f"Cannot translate a {type(efd)} to an EfdClient.")
 
@@ -85,6 +86,7 @@ async def query_efd_topic_for_night(
             results.append(result)
 
     result = pd.concat(results) if len(results) > 0 else pd.DataFrame()
+    result.index.name = "time"
 
     return result
 
--- a/schedview/collect/efd.py
+++ b/schedview/collect/efd.py
@@ -1,0 +1,139 @@
+import os
+from collections import defaultdict
+from collections.abc import Iterable
+from functools import partial
+
+import pandas as pd
+from lsst_efd_client import EfdClient
+
+from schedview.dayobs import DayObs
+
+DEFAULT_EFD = (
+    "summit_efd" if os.getenv("EXTERNAL_INSTANCE_URL", "") == "https://summit-lsp.lsst.codes" else "usdf_efd"
+)
+SAL_INDEX_GUESSES = defaultdict(partial([[]].__getitem__, 0), {"lsstcomcam": [1, 3], "latiss": [2]})
+
+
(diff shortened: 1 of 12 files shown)


ehneilsen:
Times Square is down, so here is an example output plot.
[example_timeline.html.gz](https://github.com/user-attachments/files/18324424/example_timeline.html.gz)

rhiannonlynne (approved):
I think it would be really useful to expand the documentation to make explicit 
* why there are multiple sub-classes of the timeline log plot class (by making it clear what is different about the subclasses and WHY, it helps people understand what they would be expected to add if they make new subclasses .. you can see this to some extent from the code itself, but spelling it out is helpful especially if someone is coming in with a very different model of how it ought to work in mind)
* make it clear what the "kwarg" dict is supposed to be in the multitimeline plot function. Some examples in the docstring would be helpful.

I think that currently this is testing by running against the actual EFD. This is not super great -- Merlin suggested looking into "vcr" and at first glance, this seems like it might be useful. Can you look at replacing this unit test with one using vcr? I think that can be a separate ticket. Maybe disable the tests that run against the actual EFD unless the hostname is within usdf for now? (that's pretty close to "only run when triggered by a human" which is probably infrequent enough that it's fine).

rhiannonlynne:
Here's the VCR readthedocs page - https://vcrpy.readthedocs.io/en/latest/
We use the EfdClient to do the efd queries, and I'm not 100% sure if this will confuse things or not, but eventually these are http queries so it might be just fine. 

(one final thing -- I'm not entirely sold on swapping to makeEfdClient, mostly because it fails everywhere that's not a rubin site, and it's clear that the logging team are expecting to be able to run their logging tools outside those locations .... I'm a bit surprised this runs the tests in github with that import line in efd.py even?)

ehneilsen:
> I think that currently this is testing by running against the actual EFD. This is not super great -- Merlin suggested looking into "vcr" and at first glance, this seems like it might be useful. Can you look at replacing this unit test with one using vcr? I think that can be a separate ticket. Maybe disable the tests that run against the actual EFD unless the hostname is within usdf for now? (that's pretty close to "only run when triggered by a human" which is probably infrequent enough that it's fine).

Most of the test cases do not use the EDF, but rather construct the dictionary which would be returned by the EFD in the test case code itself. The exception is `TestTimelinePlotters.test_example`, which is skipped by default. Yes, it would be more thorough to write a full mock for the EFD client class.

rhiannonlynne:
I think the thing with VCR is that you can skip most of the work, and just add the decorators then run the test somewhere where the HTTP responses can be recorded. Then later executions of the same test with the same parameters just use the recorded responses. 
This does mean that changes in the EFD *won't* be available for tests, but that would be true of a mocked EFD anyway.

ehneilsen:
> (one final thing -- I'm not entirely sold on swapping to makeEfdClient, mostly because it fails everywhere that's not a rubin site, and it's clear that the logging team are expecting to be able to run their logging tools outside those locations .... I'm a bit surprised this runs the tests in github with that import line in efd.py even?)

I tried to be sure it would work if makeEfdClient wasn't available: see the try/except structure [here](https://github.com/lsst/schedview/blob/37ec68beb67f0a5cf4dc21eec589bb6739fde50a/schedview/collect/efd.py#L20). I suppose it would still fail if `lsst.summit.utils.efdUtils` is installed, but not at an Rubin site. I can fix that easily enough if I can figure out what exception that would throw.

rhiannonlynne:
It's a RuntimeError https://github.com/lsst-sitcom/summit_utils/blob/b167aba8ce67140068bf61055731d4b1dd37044d/python/lsst/summit/utils/efdUtils.py#L471

Ok - yes, I see the try/except now .. I think I looked at the commit where it was added, and didn't see that it had changed by the end. Ok!

PR #125, opened 2025-01-23
SP-1864: Swap target to target_name in DDF constraint

rhiannonlynne:
The constraint for read_ddf_visits used "target" (old column name) instead of "target_name" (new column name). 
Many other uses of opsim are backward compatible, but it's not obvious how to make this particular use backward compatible unless you retrieve all visits and then sub-select afterwards, or query table first to check for columns .. essentially doing the work that is done in the next step of "read_opsim", and I didn't want to duplicate code.

### DIFF
--- a/schedview/collect/opsim.py
+++ b/schedview/collect/opsim.py
@@ -211,7 +211,7 @@ def read_ddf_visits(
     visits : `pandas.DataFrame`
         The visits and their parameters.
     """
-    ddf_field_names = tuple(ddf_locations().keys())
+    ddf_field_names = [f"DD:{field_name}" for field_name in ddf_locations().keys()]
     # Note that this where clause is hard-coded for target_name (v4+)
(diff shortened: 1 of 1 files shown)


ehneilsen (approved):
Looks fine to me.

rhiannonlynne:
Also - I hear your concern about it not being backward compatible. 
I thought about how to do that and didn't come up with a super satisfactory answer. 
I suppose (thinking about it again) the option I have is basically - try with this new version, and if it fails, try with the old version. But that feels kind of clunky and not future-proof (when it changes again, do we just try again)? 
Or otherwise, fetch the data first and sub-select in this section. 
But yes -- the forward/backward compatibility here feeds into part of my questions about data access. 
For now, given that we don't have a better option - I think doing the try / except maybe makes sense, and then we fix it better later.

PR #130, opened 2025-02-20
updates to align more with DM style guide

humnaawan:


ehneilsen (approved):
Sure, this is fine. I think it would be even better to use `ET` instead of either `et` (what I used initially) or `ElementTree` (what you've changed it to), because `ET` is what the python.org examples use in the [official python documentation of xml.etree](https://docs.python.org/3.11/library/xml.etree.elementtree.html).

PR #131, opened 2025-02-28
tickets/SP-1944: Add query argements to set the time to the URL for the schedview snapshot dashboard

ehneilsen:


schedview/app/scheduler_dashboard/lfa_scheduler_snapshot_dashboard.py
```diff
@@ -79,28 +92,120 @@ async def query_schedulers(self, selected_time, selected_tel):
...
+
+    @pn.depends("mjd_cut", watch=True)
+    def update_datetime_cut_from_url(self):
+        self.logger.debug("UPDATE: datetime_cut from mjd_cut")
+        datetime_cut = Time(self.mjd_cut, format="mjd")
+        if self.datetime_cut != datetime_cut.datetime:
+            self.datetime_cut = datetime_cut.datetime
+
+    @pn.depends("datetime_cut", "telescope", watch=True)
+    def get_scheduler_list(self):
```
emanehab99, line 165: May be this function should be async because it's running async task?

schedview/app/scheduler_dashboard/scheduler_dashboard_app.py
```diff
@@ -114,27 +117,23 @@ def scheduler_app(date_time=None, scheduler_pickle=None, **kwargs):
...
+        url_sync_map["telescope"] = "telescope"
 
     # Restrict files to data_directory.
     else:
         scheduler = RestrictedSchedulerSnapshotDashboard(data_dir=data_dir)
-        # data_loading_widgets = {
-        #     "widget_datetime": pn.widgets.DatetimePicker,
-        # }
+
+    pn.state.location.sync(scheduler, url_sync_map)
```
emanehab99, line 136: The reason why the test is failing here is that the dashboard app isn't running through the server (we just call the function). Hence the URL is None.

emanehab99 (changes requested):
Looks good to me @ehneilsen . I left a couple of comments those might fix the tests. 
PS. I couldn't run the dashboard on USDF. For some reason the port forwarding didn't work for me as before; give me 404! Any tips to make it work?

emanehab99 (commented):
Apologies I didn't mean to request changes. These are just comments

schedview/app/scheduler_dashboard/lfa_scheduler_snapshot_dashboard.py
ehneilsen: It creates a task, yes, but it adds it to the event loop instead of awaiting it: it does not itself need to be wrapped in a task and added to the event loop, so I don't think it should by async. (This was the point of separating `_async_get_scheduler_list` from `get_scheduler_list`.) But, maybe I'm confused?

ehneilsen: I eventually discovered that the param module has its own mechanism for adding async funcs to the event loop, so switched to that.

rhiannonlynne:
Eman's had much better comments here than I -- I don't know enough about the underlying code and how it works to really have good insight. 
But it looks fine to me. 

Although .. one question - did the URL for the snapshot ever get put in a place where it could be copy/pasted? If not .. if it shows up in the debugging window, is that somewhere that can be copy/pasted from? (the debug window would be fine ... not everyone needs to do this, but it would be nice to be able to).

rhiannonlynne (approved):

rhiannonlynne (approved):

rhiannonlynne:
> Apologies I didn't mean to request changes. These are just comments

Hi @emanehab99, I think you might need to change your review response to make it easier for @ehneilsen to merge (the "Requested changes" moves the merge to blocked). If you click the little circling arrows next to your name under reviews you can change the value in the dropdown again.

ehneilsen:
> Although .. one question - did the URL for the snapshot ever get put in a place where it could be copy/pasted? If not .. if it shows up in the debugging window, is that somewhere that can be copy/pasted from? (the debug window would be fine ... not everyone needs to do this, but it would be nice to be able to).

Changing the "Starting to load scheduler" log message currently in the debug section to include the snapshot URL is easy, just adding `+ "self.scheduler_fname` to the logging line. I'll do that now.

schedview/app/scheduler_dashboard/lfa_scheduler_snapshot_dashboard.py
emanehab99: Good to know!

emanehab99:
Glad you managed to fix it and merge the PR @rhiannonlynne

PR #135, opened 2025-03-21
SP-2039: Update workflows

rhiannonlynne:


ehneilsen (approved):
This all looks good to me.
I wonder when we should drop python 3.11 and add 3.13.

rhiannonlynne:
Good question. Itd possible telescope and site are still on py3.11? I should check.

PR #137, opened 2025-04-07
tickets/SP-2071: Update schedview & schedview_notebooks to use band instead of filter when it is available

ehneilsen:


rhiannonlynne:
Can you tell me where the use of 'filter' was failing? I thought the expected behavior was that filter would always be populated, and if band_to_filter was not set to a different value that "filter" would just match "band". 
This seems like there may have been a bug somewhere else?

ehneilsen:
> Can you tell me where the use of 'filter' was failing? I thought the expected behavior was that filter would always be populated, and if band_to_filter was not set to a different value that "filter" would just match "band".
> This seems like there may have been a bug somewhere else?

Yes, there were visits I was loading from somewhere that were missing filter entirely, but I thought we were discontinuing the use of filter, so I just fixed schedview to deal with it. I now no longer remember exactly what prompted me to do this, though.

PR #139, opened 2025-04-21
tickets/SP-2005: Add comparisons to pre-night simulations to the scheduler-focused night summary report

ehneilsen:
Includes a stand-alone sample panel app.


rhiannonlynne:
Where's the standalone sample panel app and do you have a screenshot? (might be helpful). 

This was hard for me to follow through - it kind of seems like primarily the PR reorganizes how data is retrieved and then has additional columns added (through the consolidation of lists of rubin_sim stackers and the modification of some of those stackers, including the time stamp values .. there are a lot of just very slightly different time-related stackers, and it might be worth considering if those can be down-selected or simplified?). 

I think this PR has added a comparison of "was visit X close enough to the same as visit Y" (from a different simulation), or at least updated it from previous versions .. I'd like a better understanding of what does count as the same and what doesn't so maybe some examples would be helpful?

rhiannonlynne:
I can't run the schedview_notebooks that would demonstrate this PR, and I can't very easily see what the new feature is (here is where a screenshot would be helpful, or instructions on how to run a thing easily). 
In general .. it seems ok? But I'm not really following along with what the new features are.

rhiannonlynne (approved):

PR #140, opened 2025-04-23
tickets/SP-2173: Support reading scheduler pickles with extra stuff

ehneilsen:


tests/test_scheduler_pickle.py
```diff
@@ -32,6 +32,21 @@ def test_bare_scheduler(self):
...
+            content = (scheduler, conditions, extra_content)
+            sample_path = os.path.join(data_dir, "sample_scheduler.pickle")
+            with open(sample_path, "wb") as file_io:
+                pickle.dump(content, file_io)
+
+            scheduler, conditions = read_scheduler(sample_path)
+
+        self.assertIsInstance(scheduler, CoreScheduler)
+        self.assertIsInstance(conditions, Conditions)
+
```
yoachim, line 49: Maybe want to also check that `extra_content` is able to be restored just for completeness.

yoachim (approved):
Looks good

tests/test_scheduler_pickle.py
ehneilsen: Okay, I've added an option just to get everything (but I don't use this anywhere).

PR #143, opened 2025-04-29
tickets/SP-2120: Update schedview to place legends correctly with bokeh 3.7.0 and later.

ehneilsen:


rhiannonlynne (approved):
I had some comments that ended up outdated once the force push went through - please ignore if they still show up. 
Moving the legend looks fine.

PR #147, opened 2025-06-09
tickets/SP-2290: 	 Add tools to schedview to plot outlines of the survey footprint

ehneilsen:
See the new notebook `notebooks/footprint_polygons.ipynb` for examples of what it does.

### DIFF
--- a/notebooks/footprint_polygons.ipynb
+++ b/notebooks/footprint_polygons.ipynb
@@ -101,12 +101,10 @@
    "source": [
     "psphere = Planisphere()\n",
     "asphere = ArmillarySphere()\n",
-    "msphere = MollweideMap()\n",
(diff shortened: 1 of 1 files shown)


rhiannonlynne (approved):
This is a great start - I think we might need to think about how the outlines should look for simplicity/readability, and not show all of them (and maybe set up a default set as a result). But this is good. 
I take it the outlines are not saved, but generated each time? Does it make sense to save them?

ehneilsen:
> This is a great start - I think we might need to think about how the outlines should look for simplicity/readability, and not show all of them (and maybe set up a default set as a result). But this is good. I take it the outlines are not saved, but generated each time? Does it make sense to save them?

We might eventually want to save them, but we need to be careful about how to do it: if we save them, there is a danger of the saved outlines becoming out of date without anyone noticing. I think that's a task for a separate PR.

PR #148, opened 2025-06-16
SP-2309: specify now-required extent argument is skyproj.LaeaSkyproj

ehneilsen:
Fixes issue with skyproj that is preventing accumulated depth from showing up in the nightsim


rhiannonlynne (approved):
Might not be bad to put this into rubin_sim as defaults as well.

PR #152, opened 2025-06-18
tickets/SP-2348: pass url to the call to load_consdb_visits

ehneilsen:


rhiannonlynne:
Why did you need to  pass the url to load_consdb_visits? 
I'm assuming this is because of the usdf-rsp-dev vs usdf-rsp difference, but I can't tell.  It could also support 
I also don't know how the token fits in here, if you're going to pass the url I think you should also allow to pass the token, so that the options are available to run this outside of the RSP.

rhiannonlynne (approved):

PR #172, opened 2025-08-20
add basic sv stat table

yoachim:
Add basic stats table.


ehneilsen (approved):
Only one minor comment.
A unit test would be nice.
Otherwise, looks good.

PR #174, opened 2025-09-11
tickets/SP-2608: Update so CI tests run with baseline 5.0.0

ehneilsen:


schedview/collect/opsim.py
```diff
@@ -186,11 +186,10 @@ def read_ddf_visits(
     visits : `pandas.DataFrame`
         The visits and their parameters.
     """
+    ddf_field_names = [f"DD:{field_name}" for field_name in ddf_locations().keys()]
+    ddf_field_names += [f"dd:{field_name}" for field_name in ddf_locations().keys()]
```
rhiannonlynne, line 190: The variation is in the entire field_name, not just whether it's DD or dd at the start .. plus ":" may not be the way we choose to go with this I think (":" is a special character in the camera metadata, where observation_reason passes through). 

When creating DDF field names for display, it would be fine to *add back in*  the DDF part but when querying the visits .. why not look for the field name only? 

So your query below would be like: 

```
constraint = " OR ".join([f"target_name LIKE '%{field_name}%'" for field_name in ddf_locations().keys()])
```

and the ddf_field_name you use for display can be 
```
ddf_field_names = [f"DDF {field_name}" for field_name in ddf_locations.keys()]
```

PR #200, opened 2026-04-02
tickets/SP-3038: Enhanced VisitMapBuilder controls

ehneilsen:
Add zenith button, play button, coords system selector button, and fancy controls layout.
The example in the example notebook with all the bells and whistles includes examples of using these.


rhiannonlynne (approved):
Can't wait to use the new controls.

PR #202, opened 2026-05-11
tickets/SP-3022: option to map colors to observation_reason in plot_visit_param_vs_time

ehneilsen:
The initial draft of the code was done with AI (claude code), and then edited to fix it.


rhiannonlynne (approved):
It looks like all of the observation reasons available across all of the range of simulations ends up in the list at the bottom?  The only drawback here is that it makes it a bit complicated to figure out what changed in different realizations of the different simulations. 
Still - I think this is more useful in seeing how FBS choices changed than the color-coding per filter, so let's try it!

PR #203, opened 2026-05-12
tickets/SP-2962: automatic regeneration of sample data for tests

ehneilsen:
This change stops the test suite from depending on manually refreshed sample artifacts in `schedview/data` during pytest runs.

 At test startup, `tests/conftest.py` now ensures there is a generated sample-data directory under `.pytest_cache/schedview-sample-data/<manifest-hash>`. Here, the “manifest” is the small JSON description of the generation inputs and expected outputs: Python version, `rubin_scheduler` version, requested date/duration, generator source hash, and the expected artifact filenames. That manifest is hashed to pick the cache directory name, so a change to one of those inputs automatically selects a different cache entry.

If that cache directory does not already exist, `schedview/testing/sample_data.py` runs the existing `rubin_scheduler`-based generation path and writes fresh versions of:

  - `sample_opsim.db`
  - `sample_rewards.h5`
  - `sample_scheduler.pickle.xz`

The test process is then pointed at that generated directory via `SCHEDVIEW_SAMPLE_DATA_DIR` and `SCHED_PICKLE`, so the tests use artifacts produced by the current runtime environment rather than whatever happens to be checked into `schedview/data`.

 The cache key is derived from the inputs that actually affect generation:
  - Python version
  - `rubin_scheduler` version
  - requested date/duration
  - generator source (that is, the python file that generates the test data) hash
  - expected artifact set

That means a scheduler-version change automatically yields a new cache entry, while repeated local test runs can reuse existing generated data. There is no special CI cache wiring here: in CI, if .pytest_cache is empty, the sample data is regenerated and used for that run; the packaged sample data is not used by pytest unless this bootstrap path is bypassed.

AI-assisted note: these changes were mostly generated with Claude Code, running on the azure/gpt-5.4 model, and then adjusted my hand.

### DIFF
--- a/schedview/collect/scheduler_pickle.py
+++ b/schedview/collect/scheduler_pickle.py
@@ -7,10 +7,13 @@ import lzma
 import os
 import pickle
 from collections.abc import Sequence
+from pathlib import Path
 
 from lsst.resources import ResourcePath
 from rubin_scheduler.scheduler.model_observatory import ModelObservatory
 from rubin_scheduler.scheduler.schedulers.core_scheduler import CoreScheduler
+
+from schedview.testing.sample_data import SAMPLE_DATA_DIR_ENV_VAR
 
 try:
     PICKLE_FNAME = os.environ["SCHED_PICKLE"]
@@ -124,6 +127,12 @@ def sample_pickle(base_fname="sample_scheduler.pickle.xz"):
     fname : `str`
         File name of the sample pickle.
     """
+    sample_data_dir = os.environ.get(SAMPLE_DATA_DIR_ENV_VAR)
+    if sample_data_dir is not None:
+        sample_path = Path(sample_data_dir).joinpath(base_fname)
+        if sample_path.exists():
+            return str(sample_path)
+
     root_package = __package__.split(".")[0]
 
     try:
--- a/schedview/testing/__init__.py
+++ b/schedview/testing/__init__.py
@@ -1,0 +1,21 @@
+from .sample_data import (
+    CACHE_SCHEMA_VERSION,
+    SAMPLE_DATA_DIR_ENV_VAR,
+    SAMPLE_PICKLE_ENV_VAR,
+    ensure_cached_sample_data,
+    generate_sample_data_dir,
+    get_sample_data_dir,
+    get_sample_data_path,
+    write_sample_data,
+)
+
+__all__ = [
+    "CACHE_SCHEMA_VERSION",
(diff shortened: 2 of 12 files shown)


rhiannonlynne:
This seems nice, but I'm not sure I understand why you need the sample scheduler (which was the thing which kept failing). 

In the same test that reads the pickle, you also generate a scheduler.  What was the drawback of just writing that scheduler to a temporary file pickle, testing it could be read with the standard methods, and then deleting it immediately afterward? 

As far as writing to the cache, I guess the other question I had is how it gets deleted? Or is this something the user would take care of?

rhiannonlynne (approved):
Some questions in the 'conversation' area

ehneilsen:
> This seems nice, but I'm not sure I understand why you need the sample scheduler (which was the thing which kept failing).
> 
> In the same test that reads the pickle, you also generate a scheduler. What was the drawback of just writing that scheduler to a temporary file pickle, testing it could be read with the standard methods, and then deleting it immediately afterward?

It's used several places in `test_scheduler_dashboard.py`. I suppose the pickle generation and caching code _could_ be moved in to `TestSchedulerDashboard.setUp`, but I can't see any advantages to doing that, and do see some downsides: other potential tests couldn't use it, and we couldn't use the same infrastructure used for `sample_opsim.db` or `sample_rewards.h5`.

Several other tests _do_use `sample_opsim.db` or `sample_rewards.h5`.  Differences in `rubin_scheduler` version are less likely to "break" a `sample_opsim.db` because we keep that schema fairly stable, but differences in `rubin_scheduler` are more likely to lead to meaningful changes to `sample_rewards.h5` if things like names of basis functions change. (I'm uncertain how sensitive currently existing tests are to these changes, but I think it's good in principle to make sure that there is consistency.)

ehneilsen:
> As far as writing to the cache, I guess the other question I had is how it gets deleted? Or is this something the user would take care of?

At present, it's something for the user to take care of. In CI, the whole environment is ephemeral so it shouldn't be an issue. A `git clean -d` (which is something I do occasionally anyway) will clean it up for a checked-out working directory, and that would satisfy my development process, but if you wanted an automatic clean up we'll need to figure out an appropriate condition to initiate the cleanup. The simplest might be to just delete any old ones when a new one is made, but I think there might be a good chance of deleting one that might still be useful in that case, which we might want to do if we were testing `schedview` with different versions of `rubin_scheduler` (not common, but it has happened).

PR #205, opened 2026-06-25
SP-3256: Add function to fetch dayobs-downtime summary

rhiannonlynne:
I'm not sure if this is what you had in mind or not. 
I looked at just adding this in a notebook, but the information should go into one of the other HTML markup nicely formatted pieces you had in the output notebook, so that doesn't quite work. 
This is about the best I could do for now, but I think you mentioned wanting to rationalize some of the rubin nights vs schedview fetching .. maybe this both indicates that that's a good idea but also illustrates an option?

### DIFF
--- a/schedview/collect/observatory_state.py
+++ b/schedview/collect/observatory_state.py
@@ -1,0 +1,48 @@
+import pandas as pd
+import rubin_nights.dayobs_utils as rn_dayobs
+import rubin_nights.observatory_status as obs_status
+from astropy.time import TimeDelta
+from rubin_nights.influx_query import InfluxQueryClient
+
+import schedview.clientsite
+
+from .auth import get_auth
+
+
+def get_observatory_status_summary(day_obs: int, day_obs_max: int | None) -> pd.DataFrame:
+    """Return a summary of observatory status (WEATHER, DOWNTIME, etc).
+
+    Parameters
+    ----------
+    day_obs : `int`
+        Day_obs (YYYYMMDD) to start query.
+    day_obs_max : `int` or `None`
+        Day_obs (YYYYMMDD) to end query. If None, query only the first day_obs.
+
+    Returns
+    -------
+    obs_status_summary : `pd.DataFrame`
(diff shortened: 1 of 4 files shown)


ehneilsen:
Can you add a simple pytest or unittest? You can (probably should) make it get skipped by default do it doesn't get run by github CI (see what I do in tests/test_efd.py).

rhiannonlynne:
Test added, follows your key for the environment variable to flag whether or not to run the live unit test.

rhiannonlynne:
I think these unit tests are all failing for something other than what I changed -- and I think you might have a fix over on your PR branch?

ehneilsen:
> I think these unit tests are all failing for something other than what I changed -- and I think you might have a fix over on your PR branch?

It looks like the tests are failing because the new code depends on rubin_nights, which is not currently a schedview dependency, and so is not getting installed in the CI environment.

rhiannonlynne:
The old failures were something else .. I hadn’t looked at the new failures
since my update to the current main today.
Makes sense about Rubin nights though! Ok will go fix that next.


================================
Dr R. Lynne Jones
Research Scientist with Rubin Observatory
she/hers
================================

schedview/collect/observatory_state.py
```diff
@@ -0,0 +1,48 @@
...
+    -------
+    obs_status_summary : `pd.DataFrame`
+        Dataframe with index `day_obs` and columns including `weather_down`
+        indicating the hours of weather downtime during the night.
+    """
+
+    # Make some assumptions that ought to work in an RSP environment,
+    # or user's env variables match schedview's expectations,
+    # e.g. tokenfile set in env variable ACCESS_TOKEN_FILE.
+    site = schedview.clientsite.EFD_NAME.replace("_efd", "")
```
ehneilsen, line 32: A more direct way to do this would be `schedview.clientsite.guess_site()`.

rhiannonlynne: I thought your sites were different. But also you have to account for usdf vs usdf-dev. I think that's what I was trying to accommodate here?

